PR #16, opened 2026-06-27
Cross-task GEPA model (tau2 + terminal)

SilenNaihin:
Ships world-models/cross-task/ — a GEPA env prompt optimized across BOTH tau2-bench and terminal-tasks combined, so it generalizes to API-style and shell-style benchmarks instead of overfitting to one.

Rebased onto current main (#13 reporting + #17). The benchmark-results README section that was originally here is REMOVED — its numbers came from the noisy 12-trace corpus and are superseded by the trustworthy results in #20 (66-trace corpus). This PR is now just the model artifact.

Note: cross-task fidelity should be re-measured on the expanded corpus from #20 (GEPA-research follow-up); the model here was trained on the old small corpus.

ruff + ty clean; 250 passed.


SilenNaihin:
Closing: the cross-task model here was trained on the deprecated 12-trace corpus (noisy GEPA results). The cross-task *idea* (optimize across tau2+terminal to avoid single-benchmark overfit) is worth keeping, but the artifact should be rebuilt on the 66-trace corpus from #20 — folded into the GEPA-research follow-up. Superseded by #20.

PR #32, opened 2026-06-29
Trace scaling law: fidelity vs. number of training traces

SilenNaihin:
## What

Adds a **trace scaling law** experiment to the research harness (`wmh.research`): how does world-model reconstruction fidelity grow as we feed it more *training* traces — and where does it saturate? The x-axis is the number of training traces (10 → up to the corpus size, ultimately 1000); the y-axis is held-out fidelity.

Built entirely on the existing `Ablation` framework (`conditions × seeds → mean ± std`) and reusing the canonical `optimize_prompt` / `score_prompt` — no scorer reimplementation, so the metric is comparable to `wmh eval`.

## Design (decided with @silen)

**Fixed test, growing train, 3-way split** (`wmh/research/scaling_split.py`):
- `test` and `valid` bands are pure functions of `trace_id` (a stable blake2b hash), so growing the corpus only ever enlarges the train pool — every point on the curve is scored on the *same* held-out test set.
- The train subsample is **nested** for a fixed seed (n=10 is a prefix of n=20): each step up the curve adds traces rather than resampling, isolating corpus size from sample luck. Seeds give the error bars.
- This is a **3-way train/valid/test** split (GEPA selects on `valid`, fidelity reported on a `test` set GEPA never saw) — stricter than the 2-way build/bench split, so the scaling curve isn't optimistic.

**Two curves** (`wmh/research/trace_scaling.py`, `TraceScalingAblation` sweeps `mode × count`):
- `base` — shipped `BASE_ENV_PROMPT` + RAG buffer over N traces. No GEPA: cheap, isolates retrieval-only scaling. Scout the range first.
- `gepa` — GEPA-optimized on N traces. The real "learning from more traces" curve (one GEPA run per count × seed).

Counts cap at the corpus, so the same sweep (`10,20,…,1000`) runs unchanged on today's 66-trace tau2 corpus and a future 1000-trace one.

## Surface

- `scripts/run_trace_scaling.py` — live runner (mirrors `run_seed_stability.py`). Takes a **benchmark name** (`tau-bench`, reusing its corpus + pinned judge) or a raw `--file`; `--counts` / `--modes` / `--seeds`; writes an `AblationReport` JSON. No new CLI command (keeps the surface small — research lives in `scripts/` like seed-stability).
- `docs/trace_scaling.md` + `research_directions.md` entry — design, how to run, and the path to grow the corpus toward 1000 via `tools/*-capture`.

## Extensibility

tau2 today; **terminal-tasks and swe-bench are just a different benchmark name** (corpora already committed under `examples/`). The ablation takes an ingested corpus + base prompt — nothing tau2-specific. Their corpora are smaller now, so expect short curves until generated.

## Generating more traces

This PR runs against the **current** corpora (tau2=66 → train pool ~45). The infra to reach 1000 is in place (`tools/tau2-capture/`, etc., documented); generation is a separate live-Bedrock run you trigger.

## Verification

`uv run ruff check .` ✓ · `uv run ty check` ✓ · `uv run pytest -q` → **281 passed / 6 skipped** (13 new). End-to-end dry-run of the runner on `tau-bench` with a fake provider produces a clean saturating curve across base+gepa × counts × seeds.

### DIFF
--- a/docs/benchmark_results.md
+++ b/docs/benchmark_results.md
@@ -1,9 +1,16 @@
 # Benchmark results: reproducibility
 
 The headline numbers in the README's *Benchmark results* section come from open-loop reconstruction
-fidelity (`wmh eval`) on the committed `examples/tau2-bench.otel.jsonl` corpus (66 traces / 433
-steps; telecom + airline + retail, captured from Sierra's real tau²-bench). This doc records the
-exact methodology so the numbers can be regenerated.
+fidelity (`wmh eval`) on the committed `examples/tau2-bench.otel.jsonl` corpus (telecom + airline +
+retail, captured from Sierra's real tau²-bench). This doc records the exact methodology so the
+numbers can be regenerated.
+
+> **Note (corpus grown to ~1000 traces).** The fidelity tables below were measured on an earlier
+> 66-trace / 433-step cut of this corpus. The committed corpus is now ~1000 traces (~5300 steps) —
+> see [`trace_scaling.md`](./trace_scaling.md) for how it was captured. The methodology and commands
+> here are unchanged, but the specific numbers need re-running on the larger corpus (the deterministic
+> split now yields a much bigger held-out set). Treat the tables below as the small-corpus baseline
+> until re-measured.
 
 ## Reproduce
 
--- a/docs/research_directions.md
+++ b/docs/research_directions.md
@@ -1,10 +1,18 @@
-# Research directions (not yet run)
+# Research directions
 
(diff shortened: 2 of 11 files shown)


SilenNaihin:
Superseded by #42, which landed the trace-scaling-law code (wmh/research/trace_scaling.py, scaling_split.py), the runner, the results, and the figure on the current eval-suite structure. This branch is based on the pre-#38 layout and would regress it. Closing.

PR #131, opened 2026-07-08
feat(cli): add wmh providers list for offline credential status

taneja-garvit:
Adds wmh providers list, which prints a table of each LLM backend and whether its credential env vars are set — no API calls, unlike providers verify. Reuses the same _has_credentials helper the build wizard uses.

Test mocks env vars and checks the table shows openai as configured and anthropic as not.

uv run pytest wmh/ -q: 641 passed, 10 skipped. uv run ty check: clean.


kfallah:

PR #124, opened 2026-07-07
test: fix wmh test portability on Windows

taneja-garvit:
Ran `uv run pytest wmh/ -q` locally on Windows and hit 3 failures, none related to actual logic:

- `app_test.py::test_examples_run_invokes_task_launcher` - assertion compared a hardcoded forward-slash path, fails on Windows
- `dotenv_test.py::test_upsert_env_var_appends_and_replaces` - asserts chmod 0600, which Windows doesn't support
- `dotenv_test.py::test_upsert_env_var_refuses_symlinked_env` - needs symlink creation, which requires elevated privileges on Windows

Fixed the path comparison with `Path.as_posix()`, gated the chmod check to skip on Windows, and skipped the symlink test on Windows with a reason.

Test-only change, nothing in production code touched.

`uv run pytest wmh/ -q` → 454 passed, 10 skipped, 0 failed.

PR #133, opened 2026-07-08
feat: seed demo sessions from trace state_before

taneja-garvit:
Seeds the demo replay session from the trace's initial `state_before` instead of always starting empty. `run_demo` (and the `_first_prompt` preview, which must mirror it) now pass `trace.steps[0].state_before` into `new_session(seed_state=...)` when it carries structured config or a scratchpad - closing the gap noted in `world_model._update_state`'s docstring ("structured state is left to explicit seeding/tooling").

Two tests added in `demo_test.py`: a populated seed lands in `session.state` and shows up in the step-1 prompt, and an empty `state_before` still starts from a default `EnvState()` with no behavior change.

`play` is intentionally left out of this PR -- it doesn't load a trace today, so seeding it needs a separate design decision.

Checks: `uv run ty check` clean. `uv run pytest wmh/ -q` - 640 passed, 9 skipped, with 3 pre-existing Windows failures tracked separately in #124 (unrelated to this change).

### DIFF
--- a/wmh/cli/app.py
+++ b/wmh/cli/app.py
@@ -67,7 +67,7 @@ from wmh.config import (
 )
 from wmh.engine.build import build as run_build
 from wmh.engine.build import ingest
-from wmh.engine.demo import run_demo
+from wmh.engine.demo import run_demo, seed_state_for
 from wmh.engine.eval_suites import (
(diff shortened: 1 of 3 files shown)


kfallah:

PR #140, opened 2026-07-10
Live agent sessions: interactive multi-turn pi runner + host engine

kfallah:
## What

The world-model-harness half of **live agent sessions** — running the REAL vendored pi harness **interactively** (multi-turn, steer, interrupt) inside an E2B sandbox, with tools that execute **for real** against the sandbox filesystem. The worker LLM stays host-side, so credentials never enter the sandbox (unchanged RunnerLink invariant).

This complements the eval path (#135, now on main): eval runs a *one-shot* episode whose tools are answered by the *world-model simulation*; a **session** is multi-turn over one accumulating transcript, its tools run against a **live filesystem**, and it emits a typed event stream a UI renders live. Because the host answers every `llm_request` and every `tool_request`, it is the **trusted observer** — the feed derives from the frames the host itself produced, so a compromised runner can't narrate a different story than the actions it actually took.

Rebased onto main after #135 landed. Platform-side integration (agent_sessions DB/routes/driver/web) is in experientiallabs/platform#282 and pins this branch. (Supersedes #139, which auto-closed when the `feat/real-harness-parallel-evals` base branch was deleted on merge.)

## Changes

- **`wmh/harness/pi_entry/runner_live.ts`** (new) — the in-sandbox session peer. One long-lived bare `Agent` from the champion's materialized code surfaces; startup self-check of `steer`/`abort`; `submit` bound as a benign run-ender; orphaned-toolCall repair after abort; turn-cap distinct from user interrupt; sequential tools + "all" steering. A host-side tool `is_error` is surfaced (thrown) so the agent can't reason from a failed action as success; `submit` honors the abort signal.
- **`wmh/harness/live_session.py`** (new) — `LiveSession` host engine: transport/platform-agnostic, injected `worker_fn` (metered) + `ToolExecutor` (sandbox FS or a local dir), thread-safe inbox drained on each `pump()`, typed `SessionEvent` stream, per-turn action budget, host-answered `read_skill`. An interrupt suppresses a racing `submit` event (the host owns event emission), cleared only at the runner's turn boundary.
- **New session frames** (additive; unknown frames already ignored both sides, so the eval path is byte-unchanged): host→runner `session_start` / `user_message` / `abort` / `ping`; runner→host `state` / `pong`.
- **`wmh/harness/pi_e2b.py`** — `start_live_runner(sandbox, …)` bootstraps `runner_live.ts` on a caller-owned sandbox and returns its channel; `session_entry_files()` exposes the runner source for the platform to upload.
- **`wmh/harness/e2b_sandbox.py`** — `set_timeout` + `metadata` on the sandbox slice/factory (keepalive + orphan reaping).

## Verified

- Gate: ruff + format + ty clean, **1135 passed / 18 skipped** (LiveSession frame choreography — steer/interrupt/budget/read_skill/submit-race variants; `start_live_runner` bootstrap).
- **Live, end to end against a real E2B sandbox**: `runner_live.ts` boots and handshakes, the vendored pi `Agent` runs a multi-turn task, `write_file` + `bash` execute against the sandbox filesystem, the file genuinely exists on read-back, `submit` fires. Zero model spend (scripted host-side worker).

### DIFF
--- a/wmh/harness/e2b_sandbox.py
+++ b/wmh/harness/e2b_sandbox.py
@@ -92,36 +92,6 @@ class SandboxFiles(Protocol):
     def read(self, path: str) -> str: ...
 
 
-class PtyHandle(Protocol):
-    """A running interactive PTY (e2b's pty handle): its pid identifies it for input/resize/kill."""
-
-    @property
-    def pid(self) -> int: ...
-
-
-class SandboxPty(Protocol):
-    """The `sandbox.pty` slice: an interactive terminal a user types into (live sessions only).
-
-    The eval backend never touches this — a PTY is the user's own scratch shell into a live
-    session's sandbox, streamed to the browser. `on_data` receives raw terminal bytes; `size` is
-    (rows, cols). Sliced so tests substitute a fake pty without importing the SDK.
-    """
-
-    def create(
-        self,
-        size: tuple[int, int],
-        on_data: Callable[[bytes], None],
-        *,
-        timeout: float | None = None,
-    ) -> PtyHandle: ...
-
-    def send_stdin(self, pid: int, data: bytes) -> object: ...
-
-    def resize(self, pid: int, size: tuple[int, int]) -> object: ...
-
-    def kill(self, pid: int) -> object: ...
-
-
 @runtime_checkable
 class SandboxHandle(Protocol):
     """The exact slice of `e2b.Sandbox` the harness uses, so tests substitute fakes."""
--- a/wmh/harness/e2b_sandbox.py
+++ b/wmh/harness/e2b_sandbox.py
@@ -92,6 +92,36 @@ class SandboxFiles(Protocol):
     def read(self, path: str) -> str: ...
 
 
+class PtyHandle(Protocol):
+    """A running interactive PTY (e2b's pty handle): its pid identifies it for input/resize/kill."""
+
+    @property
+    def pid(self) -> int: ...
+
+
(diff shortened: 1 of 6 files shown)


kfallah:
Addressed all review findings in `c3e004e` (rebased onto latest main; full gate green: ruff + ty + 1144 pytest / 18 skipped):

- **Skip tools while aborting** — while `_aborting`, a non-submit tool request is answered `interrupted` (is_error) without executing, so a cancelled turn no longer mutates the sandbox.
- **Tool executor exceptions escape** — `_execute_tool` is wrapped; a sandbox/FS error becomes an error `ToolOutcome` (tool_result + tool_response still sent) instead of crashing `pump()`.
- **Abort gate clears on any state** — `_on_state` clears `_aborting` only on the terminal `idle` frame (the runner always emits `state:idle` after a run, incl. aborts), so a `running` prompt-start frame can't re-enable a stale submit.
- **Manifest paths escape ./live** — session_start file keys are `path.resolve`d and rejected if they fall outside `./live` (blocks `../` traversal + absolute paths).
- **Workspace path becomes shell** — the workspace is `shlex.quote`d in `mkdir -p`.
- **Failed handshake leaves runner alive** — a no-hello handshake closes the channel (asks the runner to exit) before raising, so no node process/reader thread is orphaned.

New tests: abort-skips-tools, tool-executor-exception, running-doesn't-clear-gate, workspace-quoting, handshake-close.

PR #141, opened 2026-07-10
Harness ingest: body-map an existing agent repo into a harness

kfallah:
## What

Adds `wmh harness ingest`: turn an existing agent repo (its "body") into a `HarnessDoc` by **body mapping**. Walk the repo, include every relevant textual file as a pathful CODE surface (directory hierarchy preserved), and have an LLM write a **harnessdoc** per file describing what it is and how it serves the agent. The built document also carries an LLM overview (`prompt:core`), a `BODYMAP.md` index, and the default tool policy / loop params so it validates and renders like any other harness.

This is the wmh half of the platform "connect a GitHub repo as an agent" feature (platform PR: `feat/github-connect-agents`); the platform ingest worker calls this library in-process.

## Highlights

- **`wmh/harness/ingest.py`** (new) — `collect_repo_files` (zealous inclusion; explicit, reported skips for secrets/binaries/lockfiles/gitignored/oversized files), `body_map` (one metered LLM call per file + one overview call; a flaky reply costs one annotation, not the ingest), `build_ingest_doc` (paths sanitized to the surface safe-path rule with the original recorded in the harnessdoc, collisions disambiguated, the store's rendered paths reserved).
- **`wmh harness ingest <dir|git-url>`** — thin CLI over the library; local checkout or shallow clone; worker-role provider, metered, prints cost.
- **`Surface.doc`** — a new optional, unhashed annotation field (the per-file harnessdoc). `extra="forbid"` on `Surface`/`HarnessDoc` so a consumer pinned to an older schema fails loudly instead of silently dropping fields on round-trip.
- **Proposal-prompt elision** — `mutate.py` renders pathful code surfaces as header + harnessdoc + head only (and withholds their content hash), so a whole-repo doc does not blow up `wmh harness create` proposal prompts, and an inert body file cannot be truncate-replaced into the champion.
- Execution semantics: an ingested harness is a representation/editing substrate — no runtime runs the repo's own loop. Documented in `docs/harness_ingest.md`.

## Review

Built and then adversarially reviewed (5 lenses × verify); confirmed findings fixed in the second commit: correct gitignore semantics (whitelist files no longer collapse ingest to zero), de-fenced + length-clamped mapping output (hostile-repo hardening), final-segment path disambiguation, and reported symlink/pruned-dir skips.

## Tests / gate

`just gate` green: ruff check + ruff format --check + ty check + `pytest -q` (1083 passed, 18 skipped). New inline tests in `ingest_test.py`, `mutate_test.py`, `harness_app_test.py`; `doc_test.py` covers the annotation round-trip and `extra=forbid`.

### DIFF
--- a/docs/harness_ingest.md
+++ b/docs/harness_ingest.md
@@ -1,0 +1,52 @@
+# Harness ingest: body mapping an existing agent repo
+
+`wmh harness ingest` turns an agent you already have (a repo of prompts, loop code, tool
+definitions, and configs) into a `HarnessDoc`. We call the repo the agent's **body**, and the
+process **body mapping**: walk the body, include every relevant textual file, and write a
+**harnessdoc** per file describing what it is and how it serves the agent.
+
+```
+wmh harness ingest https://github.com/acme/support-bot --name support-bot
+wmh harness ingest ../my-agent --exclude 'docs/archive/*'
+wmh push support-bot --kind harness
+```
+
+## What the built document contains
+
+| Surface | Content |
+| --- | --- |
+| `prompt:core` | LLM-written overview of the agent + where its body lives |
+| `code:bodymap-md` (`BODYMAP.md`) | The index: every mapped file with its one-line role, plus skips with reasons |
+| `code:<slug>` per file | The file content, its real relative path in `Surface.path`, its harnessdoc in `Surface.doc` |
+| `tool_policy:main`, `param:*` | Explicit defaults so the doc renders and validates like any other |
+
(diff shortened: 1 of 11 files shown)


yassinetou9:
@kfallah Really like this direction — turning an existing agent repo into a harness via body-mapping feels like a very natural “connect GitHub as an agent” primitive. The per-file harnessdoc + overview + BODYMAP.md combination also makes the ingest result much more inspectable and edit-friendly than a single monolithic blob.

PR #155, opened 2026-07-14
Resolve chat token parameters by model

kfallah:
## Summary

- add the OpenAI-compatible output-token field to WMH's canonical model catalog
- resolve that field from `ProviderConfig` through a local catalog import
- keep GPT-5.x on `max_completion_tokens` and Kimi/DeepSeek on `max_tokens`
- carry the same contract into Waterfall's compact SDK payload and custom fallback rungs
- leave `wmh/providers/_openai_common.py` unchanged

## Why

Platform applied `chat_max_tokens_field="max_tokens"` to every Azure worker model. That is valid for Kimi and DeepSeek, but GPT-5.x rejects it with a 400 and requires `max_completion_tokens`.

The compatibility knowledge belongs to WMH's model catalog. Platform can now construct an ordinary provider config without duplicating model-specific policy.

## Validation

- `.venv/bin/ruff format --check .`
- `.venv/bin/ruff check .`
- `.venv/bin/ty check`
- `.venv/bin/pytest -q` (1229 passed, 18 skipped)

### DIFF
--- a/wmh/providers/base.py
+++ b/wmh/providers/base.py
@@ -89,7 +89,20 @@ class ProviderConfig(BaseModel):
     deployment: str | None = None  # Azure OpenAI deployment name
     api_version: str | None = None  # Azure OpenAI API version
     reasoning_effort: str | None = None  # OpenAI Responses reasoning.effort
+    # The serialized default stays stable for persisted configs. When callers do not explicitly
+    # set this field, built-in models resolve it from the canonical ProviderModel catalog.
     chat_max_tokens_field: ChatMaxTokensField = "max_completion_tokens"
+
+    def resolved_chat_max_tokens_field(self) -> ChatMaxTokensField:
+        """Return the output-token field accepted by this configured model."""
+        if "chat_max_tokens_field" in self.model_fields_set:
+            return self.chat_max_tokens_field
+
+        # Local import avoids a module cycle: the model catalog imports ProviderKind above.
+        from wmh.providers.models import resolve_provider_model
+
+        model = self.model_type or self.model
+        return resolve_provider_model(self.kind, model).chat_max_tokens_field
 
 
 @runtime_checkable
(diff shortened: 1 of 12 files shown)


wmh/providers/base.py
```diff
@@ -89,8 +89,21 @@ class ProviderConfig(BaseModel):
...
+    # set this field, built-in models resolve it from the canonical ProviderModel catalog.
     chat_max_tokens_field: ChatMaxTokensField = "max_completion_tokens"
 
+    def resolved_chat_max_tokens_field(self) -> ChatMaxTokensField:
+        """Return the output-token field accepted by this configured model."""
+        if "chat_max_tokens_field" in self.model_fields_set:
+            return self.chat_max_tokens_field
+
+        # Local import avoids a module cycle: the model catalog imports ProviderKind above.
+        from wmh.providers.models import resolve_provider_model
```
kfallah: this indicates a failure in design and is a hacky approach to doing this

kfallah: Agreed. I removed the `ProviderConfig` resolution method entirely, including the `model_fields_set` inspection and local import. Known model behavior now resolves directly from the canonical model catalog; the config field is only a fallback for unknown custom endpoints. The same contract now drives direct, structured, and Waterfall calls, with persisted-config coverage.

wmh/providers/base.py
```diff
@@ -89,8 +89,22 @@ class ProviderConfig(BaseModel):
...
+    # The serialized default stays stable for persisted configs. When callers do not explicitly
+    # set this field, built-in models resolve it from the canonical ProviderModel catalog.
     chat_max_tokens_field: ChatMaxTokensField = "max_completion_tokens"
 
+    def resolved_chat_max_tokens_field(self) -> ChatMaxTokensField:
+        """Return the output-token field accepted by this configured model."""
+        # Local import avoids a module cycle: the model catalog imports ProviderKind above.
+        from wmh.providers.models import resolve_chat_max_tokens_field
+
+        model = self.model_type or self.model
```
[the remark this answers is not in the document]
kfallah: An opaque Azure deployment name is not invertible to a canonical model, so WMH cannot safely migrate `customer-kimi-deployment` without external metadata; guessing from the deployment string would reintroduce the heuristic policy this PR removes. `model_type` is the migration field, and Platform already supplies it for every Azure agent config. Historical WMH configs store the canonical model in `model`, which the resolver still handles, while truly custom opaque configs retain `chat_max_tokens_field` as their explicit fallback. The persisted known-model round trip is covered by the new regression.

PR #167, opened 2026-07-18
feat(serve): rate-limited public play + catalog replay precompute

SilenNaihin:
## What

The backend half of the platform's unauthenticated world-model catalog. Two additions to the wmh (engine/backend) side that the platform frontend connects to.

### Public (anonymous) play on `wmh serve`
Logged-out visitors can step a public model live. That spends real provider tokens with no account behind it, so it runs through a limiter (`wmh/serving/public_play.py`):
- a **global step ceiling** that bounds total spend deterministically (each step is at most one model call),
- a **per-session step cap**, and a **session-count cap**.

Enabled with `wmh serve --public-play [--public-max-steps N]`, which adds capped routes `POST /public/world_models/{name}/sessions` and `POST /public/sessions/{id}/step` (unenriched sessions; typed refusal -> HTTP 429). Without the flag the routes 404. The platform proxies these for anon play.

### Catalog replay precompute
`packages/environment-capture/scripts/precompute_public_scenarios.py` samples each public model's indexed steps and runs `step_open_loop` to record action + golden observation vs the model's predicted observation, plus a few suggested actions. Run once; the output is vendored into the platform build so replays never recompute. Headline fidelity shown in the UI stays the model's official held-out score (from its card), never the self-retrieval number.

## Verification

- `ruff check .` + `ruff format --check` clean; `ty check` clean on the diff (one pre-existing unrelated diagnostic in `app_test.py`).
- `pytest wmh/serving wmh/cli` -> 281 passed. New tests: limiter ceilings, and serving routes (session, step, per-session/global 429 ceilings, disabled -> 404).
- **Live**: `wmh serve --public-play --public-max-steps 2` -> open session -> step (200, real Bedrock observation) -> step (200) -> step (429 at the ceiling).

Pairs with the platform PR that adds the anon proxy routes + the catalog UI.

### DIFF
--- a/packages/environment-capture/scripts/precompute_public_scenarios.py
+++ b/packages/environment-capture/scripts/precompute_public_scenarios.py
@@ -1,0 +1,124 @@
+"""Precompute example open-loop reconstructions for the public world models.
+
+The platform's unauthenticated catalog shows, per model, a few hard-coded "reconstruction"
+scenarios: a recorded action, the real observation the environment produced, and the world
+model's predicted observation. Recomputing these live on every page view would be slow and
+costly, so we run them once here and vendor the result into the platform frontend build.
+
+Each step's golden data comes from the model's own indexed corpus (`sample_steps`), and the
+prediction from `step_open_loop` (teacher-forced: predict, then advance from ground truth). The
+headline fidelity shown in the UI is the model's official held-out score from its card, not
+recomputed here, so self-retrieval on indexed steps never inflates the number a visitor sees.
+
+Run once, with the models' serve provider available (Bedrock):
+
+    uv run python packages/environment-capture/scripts/precompute_public_scenarios.py \
+        --out ../platform-public-catalog/apps/web/data/public-scenarios.json
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+from pathlib import Path
+
+from wmh.config.card import load_card
+from wmh.core.render import render_action
+from wmh.engine.loader import load_world_model
+
+MODELS_GLOB = "*/models/*"
+SCENARIOS_PER_MODEL = 2
+STEPS_PER_SCENARIO = 3
+OBSERVATION_CLIP = 900
+
+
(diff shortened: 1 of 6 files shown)


SilenNaihin:
## Note on the CodeQL check (pre-existing alert, not from this PR)

CodeQL reports 1 critical `py/full-ssrf` at `wmh/engine/grounding.py:115`. This is **not introduced by this PR**:

- `wmh/engine/grounding.py` is **not in this diff** (byte-identical to `main`).
- The same alert (**#8, `py/full-ssrf`, `grounding.py:115`**) is **already open on `refs/heads/main`** — it surfaces here only through CodeQL's PR-vs-base attribution.
- The new anonymous play surface **cannot reach** grounding: every public model is RAG-only (no `max_fidelity`/grounding), so no `/public/...` path invokes it.
- The code is already SSRF-guarded (`_assert_public_http_url` on the URL and on every redirect hop; the docstring discusses the residual DNS-rebinding residue), which CodeQL's taint tracking does not recognize.

The actual analysis jobs (Analyze python/js/actions) all pass. Dismissing or accepting alert #8 is a repo-security decision on `main`, independent of this PR. Happy to open a separate hardening PR for grounding if the team wants the alert cleared at the source.

SilenNaihin:
Closing as superseded. This PR had two halves; we've kept the valuable one and dropped the rest:

- **Catalog replay precompute** (`precompute_public_scenarios.py`) — SALVAGED to main in #269 (standalone, depends only on stable engine APIs). It's the generator for the platform's vendored `public-scenarios.json`, so the catalog replay data stays regenerable.
- **`wmh serve --public-play` rate-limited mode** (`public_play.py`, the CLI flag, server wiring) — DROPPED. The platform's public play does **not** go through `wmh serve`; it runs on the explabs backend (same path as authed play), gated by a validated-email hard-gate + per-IP/per-email rate limits at the platform layer (platform PR #363). So this half has no consumer.

Nothing missed — the precompute logic lives on in #269.

PR #171, opened 2026-07-18
fix(engine): prevent trace state corruption by deep copying seed_state

### LINKED ISSUE #169: bug: `new_session` corrupts `seed_state` traces in memory due to missing deep copy
### Summary

In `wmh/engine/world_model.py`, the `new_session` method accepts an optional `seed_state` (an `EnvState` Pydantic model) and assigns it directly to the new session's state:

```python
session = Session(
    id=uuid.uuid4().hex,
    task=task,
    state=seed_state or EnvState(),
    enrich=enrich,
)
```

Because Pydantic does **not** deep-copy the object during this assignment, `session.state` holds a direct reference to the provided `seed_state` (which commonly originates from a trace loaded into memory, e.g. `trace.steps[0].state_before`).

As the session progresses, `_update_state` mutates `session.state` in-place (for example, appending notes to `scratchpad` or updating `structured.profile`). Since both objects share the same reference, these mutations also modify the original trace state in memory.

As a result, repeatedly evaluating the same trace causes the initial state to become progressively polluted, breaking episode isolation and producing incorrect evaluation behavior.

---

## Steps to Reproduce

1. Load a trace object into memory.
2. Call `world_model.new_session(seed_state=trace.steps[0].state_before)` multiple times in a loop.
3. Generate a few steps in each session.
4. Inspect `trace.steps[0].state_before.scratchpad`.

### Actual Behavior

The original trace state is mutated in-place. The `scratchpad` accumulates notes from every session run, causing subsequent evaluations to start from an already modified state.

### Expected Behavior

Each session should receive an independent copy of the provided `seed_state`. The original trace object should remain unchanged regardless of how many sessions are created from it.

---

## Root Cause

`seed_state` is assigned directly to `Session.state`, resulting in shared object references:

```python
state=seed_state or EnvState()
```

Subsequent in-place mutations performed by `_update_state` therefore modify both the session state and the original trace state.

---

## Suggested Fix

Deep-copy the supplied `seed_state` before assigning it to the session:

```python
session = Session(
    id=uuid.uuid4().hex,
    task=task,
    state=seed_state.model_copy(deep=True) if seed_state else EnvState(),
    enrich=enrich,
)
```

Adityakk9031:
Fixes #169
# Description

## Summary

Fixes a critical issue in WorldModel.new_session where seeding a session with an existing state caused in-place mutations to leak back into the original trace object in memory.

## Description

In wmh/engine/world_model.py, the new_session method accepts an optional seed_state (an EnvState Pydantic model) and assigns it directly to the new session's state. Because Pydantic does not deep-copy the object during direct assignment, session.state holds a shared reference to the passed seed_state (which typically comes from a trace loaded into memory, e.g., trace.steps[0].state_before).

As the session advances, _update_state mutates session.state in-place (appending notes to the scratchpad or updating the profile). Because of the shared reference, this permanently corrupted the in-memory golden trace object. If an evaluation loops over the same trace multiple times, subsequent evaluations would start with a polluted initial state containing the accumulated scratchpad notes from previous runs, destroying episode isolation.

This PR fixes the issue by explicitly deep-copying the seed state before assigning it to the session using `seed_state.model_copy(deep=True)`.

## Checklist

- [x] Ensure session.state holds an independent deep copy of seed_state.
- [x] Prevented cross-episode state leakage and trace object pollution.

### DIFF
--- a/wmh/engine/world_model.py
+++ b/wmh/engine/world_model.py
@@ -183,9 +183,9 @@ class WorldModel:
         required for evaluation rollouts, whose PREDICTED observations must not become demos for
         later rollouts (order-dependent, self-reinforcing scores otherwise)."""
         session = Session(
-            id=uuid.uuid4().hex, 
-            task=task, 
-            state=seed_state.model_copy(deep=True) if seed_state else EnvState(), 
+            id=uuid.uuid4().hex,
+            task=task,
+            state=seed_state.model_copy(deep=True) if seed_state else EnvState(),
             enrich=enrich
         )
         self._sessions[session.id] = session
--- a/wmh/engine/world_model.py
+++ b/wmh/engine/world_model.py
@@ -183,7 +183,10 @@ class WorldModel:
         required for evaluation rollouts, whose PREDICTED observations must not become demos for
         later rollouts (order-dependent, self-reinforcing scores otherwise)."""
         session = Session(
-            id=uuid.uuid4().hex, task=task, state=seed_state or EnvState(), enrich=enrich
+            id=uuid.uuid4().hex, 
+            task=task, 
+            state=seed_state.model_copy(deep=True) if seed_state else EnvState(), 
+            enrich=enrich
         )
         self._sessions[session.id] = session
         tracker = RunTracker(run_id=session.id, kind="serve")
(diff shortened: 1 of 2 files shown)


Adityakk9031:
@SilenNaihin have a look

Adityakk9031:
@kfallah have a look

kfallah:

PR #213, opened 2026-07-20
fix(providers): accept request-bound Bedrock client tool type

kfallah:
## Scope

This branch explored Bedrock Converse compatibility for request-bound client tools. The reusable core accepts `toolUse.type = "tool_use"` only when the tool was advertised in the corresponding request and preserves strict validation for all other response shapes.

## Disposition

The branch also propagates broader response metadata through unrelated layers. It is closed without merge, and the narrow provider correction will be rebuilt on current main.


kfallah:
Closing because this branch is superseded and is not intended for merge. Exact head preserved: 9596b9fc8951419d06b23960e5719a62535d058e. No branch was deleted.

PR #256, opened 2026-07-25
Routing research: benchmark adapters, method variants, ablation grid, results dashboard

SilenNaihin:
# Routing research: what we built, what we tried, what actually works

This PR is the research layer behind the model-routing endpoint. The production half (the model pool, streaming, the OpenAI-compatible endpoint, the policy format it serves) already merged in #248. This PR contains everything we used to figure out WHICH routing method deserves to be served: benchmark adapters, every method variant, the evaluation harness, the results dashboard, and the findings docs.

TLDR: after ~1,500 method-configurations across 5 method families, exactly one query-side router earns promotion: a kNN selector with a statistical safety guard, worth about +1 point of accuracy at 27% lower cost on a large real benchmark. Everything fancier (learned models, clustering, IRT) loses to it. The bigger accuracy prizes are only reachable by looking at the model's ANSWER (not just the question), which is where next steps point.

## The setup, in one minute

We have a pool of 9 models (Claude Fable/Sonnet/Haiku/Opus, GPT-5.5, GPT-5.4-mini, DeepSeek-V4-Pro, Kimi-K2.6, GLM-5.2) with real per-token prices. A "router" decides, per request, which pool model answers. To compare routers cheaply, we precompute an outcome matrix: run every model on every task once, store reward + cost + latency + the raw replies. Then any router can be scored offline by lookup, no new API calls. This is the same trick RouterBench introduced.

- RouterBench: https://arxiv.org/abs/2403.12031 - a big precomputed grid (36k prompts x 11 models, 2023-era) of "which model got which task right, at what cost". Lets you benchmark a router without paying to re-run models. We use their data as a sanity check AND rebuilt the same grid with our own 9 models on their prompts (called "ours9" below, n=1,199 tasks, exact-match graded).
- LLMRouterBench: https://arxiv.org/abs/2601.07206 - same idea, 13 flagship 2025 models. We used it as a second validation set (and caught a data leak in the process, see methodology).

We also built 10 matrices on our own agentic corpora (tau-bench, swe-bench, financebench, terminal tasks, etc.). Important caveat: those episodes run against learned WORLD MODELS (an LLM simulating the environment) with an LLM judge scoring the outcome, not against the real benchmark harnesses. That adds noise, which we measured instead of guessing (see methodology). They are also small: 25 scenarios each, so ~7 test scenarios per split. The real-benchmark ours9 matrix is where statistical claims live; the world-model corpora are candidate-generators.

Routed-away fraction and accuracy against the confidence threshold z

*The shipped contract in one picture: fable-5 answers by default and the router only overrides it when the evidence clears a confidence bar. Turning that bar up sends far fewer requests away (69% down to 31%) while accuracy barely moves, and every setting stays about 5 points clear of running fable-5 alone. Two panels rather than two y-axes on one frame, so neither scale is chosen to flatter the other.*

## Methodology: how we keep ourselves honest

- Fit/test splits with 5 different split seeds. The router only ever sees the fit side. Every comparison is PAIRED BY SEED: router accuracy minus the best-single-model accuracy on the same seed's test set. A win means "positive average delta, wins on at least 4 of 5 seeds, and the delta is bigger than its own spread".
- Best-single baseline: the strongest single model, discovered on the fit split (not assumed). A router that cannot beat "just always use the best model" is a cost knob at most.
- Guards: routers only deviate from the baseline when in-cluster/in-neighborhood evidence clears a statistical bar; otherwise they revert. We verified this behaves correctly on "monopoly" corpora where one model dominates: every guarded router collapses exactly to the baseline there.
- Leak checks: fit/test ID overlap, and duplicate-task-text detection. This caught a fake +12pt IRT win on LLMRouterBench (327 of 468 test tasks appeared verbatim in the fit set).
- Shuffled-label controls: retrain the router on permuted rewards; it must collapse to chance. This killed one early "champion" and exposed that lexical (trigram) embeddings smuggle reply-length signal into controls; only semantic embeddings give clean controls.
- Judge noise, measured: we re-judged identical episodes 3x with the pinned judge. 15-17% of episode-to-episode reward variance is judge inconsistency; the rest is real behavior variance. So world-model results get discounted by that, quantitatively.
- Power rule: any claim needs 3+ seeds AND 30+ test scenarios. Below that a result can only be a "candidate", never a win. This rule is enforced by the dashboard itself.

Groups with signal against median test scenarios per seed

*Why that power rule bites: almost every benchmark section we have sits in the shaded zone under 30 test scenarios per seed, where even a large delta can only ever be a candidate. The one section with a real test set carries nearly all of the confirmed signal. This is the argument for the scaled capture round, not a claim that the small corpora are uninformative.*

## What we tried, method by method

Query-side methods (look at the QUESTION, pick a model):

- Avengers: https://arxiv.org/abs/2505.19797 - embed each task, cluster them, keep a per-cluster scoreboard of which model wins there. New request goes to its nearest clusters' champion. We replicated it faithfully (same kmeans, same scoring). Verdict: works on 2023-model pools where models genuinely specialize; on our 2026 pool it is a cost knob, not an accuracy win.
- Avengers-Pro / Hybrid LLM cost knob: https://arxiv.org/abs/2508.12631, https://arxiv.org/abs/2404.14618 - fit once, then slide a single cost-vs-quality dial without refitting. We implemented it. Verdict: real and useful (-27% to -90% cost for chosen accuracy trade), but DANGEROUS on corpora where one model dominates (-37pt on gaia2), because sliding toward cheap assumes a cheap substitute exists.
- ProxRouter: https://arxiv.org/abs/2510.09852 - reweight cluster evidence with an exponential tilt so outlier queries do not get routed on thin evidence. Verdict: robustness tool, not an accuracy win; its faithful version converged to the same method as our champion.
- IRT / IrtNet: https://arxiv.org/abs/2510.00844 - model each LLM with a learned "ability" number and each task with a "difficulty" (like standardized-test theory, 2PL IRT), with an MLP turning the task embedding into difficulty space. This is the "MLP feature extractor" idea. Verdict: killed, by its own research chat. The head learns ability and difficulty but almost none of the model-x-task INTERACTION, and routing accuracy lives entirely in the interaction (~0.01 nats per query of usable signal, measured). Lost 0/5 seeds to the kNN champion.
- JiSi: https://arxiv.org/abs/2601.01330 - kNN over past tasks plus a second routing pass on the model's response. Verdict: the response-side second pass added nothing (and its offline variant peeks at test data, so it cannot be deployed); the kNN core survived and evolved into our champion. Note: 0 citations, and the authorship attribution in the paper trail did not check out; we treat it as a direction, not an authority.
- THE CHAMPION - kNN + relative threshold + statistical guard: embed the request (text-embedding-3-large), find its nearest fit-set neighbors above a relative similarity bar (95% of the kth best), read which models actually won on those neighbors (distance-weighted), and only deviate from best-single when the evidence clears a paired standard-error test (z=0.5). Verdict: +1.04pt +- 0.6 at -27% cost, 5 of 5 seeds, on the real ours9 matrix. Best guarded result across every family, only method positive on held-out-cluster generalization, structurally strong on repeat traffic. We controlled for the embedding (giving other methods the same embedder does not save them), so the win is the method.
- Learning-to-defer: https://arxiv.org/abs/2006.01862 and https://arxiv.org/abs/2202.03673 - treat "route vs fall back to the safe model" as a formal deferral problem with a loss function that is actually consistent. Verdict: negative on financebench; one interesting +9.5pt candidate on continual-learning, too small a test set to trust yet.
- Conformal risk control: https://arxiv.org/abs/2208.02814 and https://arxiv.org/abs/2110.01052 - replace hand-tuned guard thresholds with a calibrated threshold that carries a finite-sample statistical guarantee. Verdict: works on iid splits; the guarantee measurably breaks under task drift (1 of 5 seeds violates its certificate), because drift breaks the exchangeability assumption. Now we know, with numbers.
- Matrix factorization routers (RouteLLM https://arxiv.org/abs/2406.18665, EmbedLLM https://arxiv.org/abs/2410.02223): killed before building - MF exists to fill in MISSING entries, and our matrices are 100% dense with only 9 models; there is nothing to impute.

Response-side methods (look at the ANSWER too):

- Best-of-2 with free signals: run a model twice, keep the better episode using zero-cost features (did it finish, how many steps). Finding worth remembering: the naive version was ANTI-correlated because of a Simpson's paradox - across tasks, more steps means a harder task and a lower score, but WITHIN a task, the attempt that did more work is the better one. Sign flips. The corrected selector picks the better episode 66-68% of the time. Verdict: real signal, but paying 2x for every request never clears the cost bar; the guard correctly declines it everywhere.
- Distilled verifier (Zooter idea: https://arxiv.org/abs/2311.08692) - train a tiny ridge model on reply embeddings to predict the judge's score, then use it as a free quality checker. Verdict: beats the free signals decisively (+0.096 selection accuracy, 5/5 seeds, clean shuffled control) but still not enough to make pay-2x-for-everything worthwhile (58% of the oracle ceiling, our pre-registered bar was 60%, so best-of-2 stays dead). The verifier itself is validated and kept.
- Cascades (FrugalGPT https://arxiv.org/abs/2305.05176, AutoMix https://arxiv.org/abs/2310.12963, and the deferral theory in https://arxiv.org/abs/2307.02764) - answer with the cheap model, have the verifier check the reply, escalate to the strong model only when it looks weak. You pay 1x plus an escalation fraction instead of 2x on everything. Verdict: NOT BUILT YET - this is the top-ranked next step, because it is the only family whose economics fit the verifier's measured skill, and the accuracy headroom (e.g. financebench, where the best possible router gains +0.39) is provably unreachable from the question alone.

## The honest scoreboard

Out of 1,534 seed-aggregated method-configurations on the dashboard (6,875 runs, recomputed 2026-07-26 as the scaled captures land):

- 72 pass the strict win rule - 66 on real-benchmark ours9 (the kNN champion family and its knob settings), 3 on repeat-traffic, 2 on held-out-cluster ours9, 1 on pooled OOD.
- 29 are "promising candidates" on small test sets (led by +18pt on tau-bench, n=7 test scenarios - exciting but unproven).
- 548 tie their baseline, 66 lose to it, and 532 are IDENTICAL to it on every seed (guards that never fired - i.e. the safety mechanism doing its job on corpora with nothing to win).

Cost against accuracy on the ours9 matrix, by method family

*The headline result. Each family's cost knob traces one curve from cheap-and-worse to dear-and-better; the promoted champion (guarded kNN) is the teal dot sitting above AND to the left of the best single model, which is the only quadrant that matters. Note fable-5 alone is both dearer and worse than routing. Whiskers are +-1 sd across 5 seeds.*

Count of method-configurations by verdict

*The same evidence without the cherry-picking: of everything we ran, a small minority beats the baseline, a third is literally identical to it because the guard never fired, and more configurations lose than win. The two grey bars are the honest majority.*

*Figures regenerate from the run records with `uv run --with matplotlib python .agents/scripts/render_routing_figures.py`; the counts above come from that same run, so they move when new captures land.*

## Per-niche optimization, not pooled

Optimizing over the pooled mix of every corpus at once is retired: routing flatlines there, because a pool-wide router has to be right about which model wins on a financial-document question and a customer-service dialogue at the same time, and the honest answer differs per niche. So each niche gets its own endpoint, and that endpoint's first win is not a clever router at all - it is discovering which single model actually owns that niche, which is fable-5 on financebench but sonnet-5 on tau-bench, where the global fallback fable-5 costs 2.8x as much and scores lower. What is left to win per niche is the response-side prize: the cascade ceiling below is worth +7.3pt at 24% lower cost on financebench, and closing the gap between that oracle ceiling and the deployable attempts is the next round's job.

Per-niche cost against reward on two scaled niches

*Two niches, two different answers. Gray dots are the nine pool models on the whole niche; the black diamond is the niche's own best single model. Coloured points are routers and cascades on held-out splits, so they are paired against the dashed line (the same anchor model on those splits), not against the diamond. Left: the deployable frontier today just is the niche's best model, and the red star is the cascade's oracle ceiling - an upper bound that reads the reward, never shippable. Right: routing adds nothing yet (every guarded config lands exactly on the anchor), so the win is the model-discovery step itself.*

One policy artifact covers the whole curve rather than one point on it: z and floor_q are read at serve time, so moving between the quality-max and balanced operating points is a config change with no refit, while lam is a fit-time knob and does require refitting the rank scoreboards. Speed comes along with the cost saving instead of trading against it, because on this pool per-call cost and per-call p50 latency correlate at r = 0.69 (cheap models are mostly the fast ones: haiku-4-5 is $0.00024 at 0.65s where fable-5 is $0.00362 at 3.31s), which is why the cost-aggressive operating point is also the fastest one.

## The production config, benchmark by benchmark

Every row is the promoted guarded kNN champion measured on that cohort's held-out splits, paired by seed against the best single model discovered on the fit side. The rightmost column names the exact run row used, because the shipped contract (adaptive neighbourhood + novelty floor + statistical guard at z=0.5) was not swept on every cohort and the closest available config differs.

| niche | discovered best-single | baseline acc | champion acc +- sd | dAcc (paired) | dcost | dp50 | seeds | n/seed | verdict | config row used |
|---|---|---|---|---|---|---|---|---|---|---|
| **ours9** (real benchmark, full power) | gpt-5.5 | 0.9557 | 0.9662 +- 0.0029 | +1.04pt +- 0.70 | -26% | -0.01s | 5 | 356 | **BEATS** | `r1-knn-adapt3-oai` |
| financebench-s80 (scaled niche) | fable-5 | 0.5540 | 0.5413 +- 0.0432 | **-1.27pt +- 2.84** | -3% | -0.02s | 5 | 24 | ties | `r3b-floorq0.5` |
| tau-bench-s80 (scaled niche) | fable-5/kimi-k2.6/sonnet-5 | 0.5092 | 0.5092 +- 0.0548 | abstains (= baseline) | +0% | +0.00s | 5 | 40 | abstains (= baseline) | `r3b-floorq0.5` |
| **tau-bench-real** (sim-to-real transfer) | fable-5 | 0.7625 | 0.7604 +- 0.0255 | **-0.21pt +- 2.55** | -34% | -0.78s | 6 | 40 | ties | `r3b-real-champ-floor` |
| tau-bench (25-scen wm) | fable-5/kimi-k2.6 | 0.3993 | 0.3993 +- 0.1192 | abstains (= baseline) | -3% | +0.01s | 5 | 7 | abstains (= baseline) | `r1-knn-adapt3-oai` |
| bird-sql (25-scen wm) | sonnet-5 | 0.9557 | 0.9557 +- 0.0511 | abstains (= baseline) | +0% | +0.00s | 5 | 7 | abstains (= baseline) | `r1-knn-adapt3-oai` |
| continual-learning (25-scen wm) | fable-5/sonnet-5 | 0.7386 | 0.7493 +- 0.0559 | +1.07pt +- 2.40 | -14% | -0.15s | 5 | 7 | ties | `r1-knn-adapt3-oai` |
| crmarena (25-scen wm) | fable-5/kimi-k2.6/opus-4-8 | 0.1007 | 0.1007 +- 0.0265 | abstains (= baseline) | +0% | +0.00s | 5 | 7 | abstains (= baseline) | `r1-knn-adapt3-oai` |
| dabstep (25-scen wm) | fable-5 | 0.2864 | 0.3036 +- 0.0784 | +1.71pt +- 3.10 | -6% | -0.24s | 5 | 7 | ties | `r1-knn-adapt3-oai` |
| financebench (25-scen wm) | fable-5/opus-4-8 | 0.5007 | 0.5007 +- 0.1013 | abstains (= baseline) | +0% | +0.00s | 5 | 7 | abstains (= baseline) | `r1-knn-adapt3-oai` |
| terminal-tasks (25-scen wm) | fable-5/sonnet-5 | 0.9436 | 0.9436 +- 0.0549 | abstains (= baseline) | +0% | +0.00s | 5 | 7 | abstains (= baseline) | `r1-knn-adapt3-oai` |
| swe-bench (25-scen wm) | fable-5 | 0.3936 | 0.3936 +- 0.1113 | abstains (= baseline) | +0% | +0.00s | 5 | 7 | abstains (= baseline) | `r1-knn-adapt3-oai` |
| wm-gaia2 (25-scen wm) | fable-5 | 0.3775 | 0.3775 +- 0.0695 | abstains (= baseline) | +0% | +0.00s | 5 | 6 | abstains (= baseline) | `r1-knn-adapt3-oai` |
| wm-tau-telecom (25-scen wm) | gpt-5.4-mini | 0.2650 | 0.2650 +- 0.0487 | abstains (= baseline) | +0% | +0.00s | 5 | 1 | abstains (= baseline) | `r1-knn-adapt3-oai` |
| **wm-all** (SAFETY row, never an optimization target) | fable-5 | 0.5255 | 0.5202 +- 0.0548 | **-0.52pt +- 0.72** | -0% | -0.04s | 5 | 63 | ties | `r1-knn-adapt3-oai` |

Reading notes. "abstains (= baseline)" means the guard never fired on that cohort, so the router served the baseline model on every scenario and scored identically by construction - that is the safety mechanism working, not a measurement. The negative rows are bolded on purpose: the champion loses 1.27pt on financebench-s80 and 0.52pt on wm-all, both inside the seed spread but both real enough to keep on the page. The "discovered best-single" column is the fit-split winner and can differ per seed (tau-bench-s80 shows three names for that reason), which is not the same thing as the matrix-wide winner drawn as the diamond in the figure above. The sim-to-real row is the one to watch: a router fitted on the tau-bench-s80 simulation transfers to the real tau-bench harness as a tie (-0.21pt) at 34% lower cost and 0.78s lower p50, so simulation-fitted routing did not fall over on contact with reality, but it did not win either. On the wm-family niches the pending safety amendment is the global-agreement gate from r3's autopsy.

## Operating points: the dial, not a single answer

Cost, accuracy and latency operating points on ours9

*The frontier is dialable per endpoint, so the question is not "does routing work" but "which point do you want on this niche". Any point at or to the right of the anchors is reachable today with the shipped knobs - z and floor_q at serve time, lam at fit time - spanning +1.14pt at 14% cheaper through to a deliberate 4.97pt quality sacrifice for 86% cheaper and 41% faster. Points ABOVE the top of this curve are not reachable by choosing a model per question at all; those need the cascade, whose oracle ceiling is the red star in the per-niche figure.*

## How to run the dashboard

From the repo root on this branch:

    uv run python .agents/scripts/build_dashboard.py
    open .wmh/evals/dashboard.html

The builder reads run records from the shared research data root (`~/Desktop/Projects/wmh-routing-data/runs/*.jsonl`). How to read the page:

- Headline tiles: one per benchmark section, only shown when a full-power result exists. Each tile carries accuracy +- spread, the paired delta vs best-single, the seed win count (e.g. 5/5), and a verdict badge (BEATS / ties / WORSE).
- Candidates strip: results on small test sets, badged "promising (small test set, n=...)" - these are the shopping list for the next data-collection round, not wins.
- Charts: one point per method-configuration (averaged over seeds, error bars = spread across seeds), log-x cost, colored by family (black diamond = best single model, teal = kNN retrieval, blue = clustering, purple = IRT, amber = prox variants, red = multi-call). Filled marker = statistically solid, hollow = within noise. A cost-knob sweep draws as one connected curve.
- Every group expands to its per-seed rows, labeled "single seed, not a result".

## Next steps

1. Promote the kNN champion into `wmh/optimize` behind the policy seam so the endpoint from #248 can serve it (one open adjudication first: whether the one competing configuration survives a calibrated guard).
2. Scaled data round (~$450 total): re-capture the two clean headroom corpora at 80 scenarios / 16 steps (fixes the n=7 problem that blocks every candidate), PLUS one matrix against the REAL tau-bench harness (no world model, no judge) to measure how well world-model-fitted routing transfers to reality.
3. Build the cascade: cheap model answers, distilled verifier checks, escalate on doubt. Pre-registered success bar before it ships, same as everything else here.
4. Optional spectrum-closer: an RBF-kernel SVM sits between clustering and kNN on the local-vs-global spectrum and is the one point not sampled; $0 to run through the existing harness.

Full findings ledger and per-round history: `.agents/docs/research/routing-method-tiers.md` and `routing-lit-review.md` in this PR.

## Where things stand (final update for this PR)

**The best result of the program, now shipped:** the guarded kNN router is production code (PR #259, merged). On the large real benchmark it beats the best single model by **+1.04 points while cutting cost 27%, on 5 of 5 seeds**, and the production implementation reproduces the research numbers to four decimal places. It serves under a conservative contract: fall back to fable-5 unless the statistical evidence for routing away is strong, with a confidence dial (tightening it to z=1.0 routes less AND scores slightly higher, at a smaller cost saving). Hardening from the specialist rounds is folded in: the neighbor budget scales to the data, and a novelty floor abstains to the fallback on requests that look unlike anything in the fit data (this closes a measured weakness where the router dipped below baseline under task drift).

**The sim-to-real test happened** (a real tau-bench matrix: 9 models x 40 tasks x 2 trials against the actual benchmark, no world model, no LLM judge):
- Model level: reality and simulation agree on the champion (fable-5 is genuinely best on real traffic, so the fallback contract is exactly right), but only moderately on the middle of the field (rank correlation 0.60), and the simulation buries one model entirely (glm-5.2: top-3 real, bottom-3 simulated).
- Decision level: routing fitted on the old 25-scenario simulated data is a wash-to-slight-loss on real accuracy (the simulator compresses model differences, so the guard fires on evidence that is not there) - but the cost savings survive (parity at up to -71% cost). The improved 16-step captures move transfer to roughly neutral, with the first genuinely positive real-accuracy routing rows.
- Consequence: simulated matrices are a cheap factory with a known distortion. One research chat is now building a sim-vs-real disagreement benchmark (with a held-out split) and attacking the distortion directly.

**The cascade found its ceiling and exhausted its free options:** a perfect escalation decision is worth +3 to +7 points at 17-34% LOWER cost (confirmed on two datasets), but every zero-cost escalation signal measurably fails. The one remaining design from the literature - paying a small LLM to check the cheap reply before escalating - is being tested now with a pre-registered pass/fail bar.

**Next phase, one line each:** push the paid-verifier cascade (the biggest open prize); a citation-screened hunt for the next method family that fits our measured constraints (tiny fit sets, noisy rewards, recurring traffic); close the sim-to-real gap empirically (fix why the simulator buries mid-field models, holdout-validated). Figures and the census above are regenerated from the live run files as of this update; exact counts move as new experiments land.

## Maintenance note (2026-07-27)

Rebuilt on wmo main (the package rename): all module paths and imports are wmo now, and this PR ABSORBS the three specialist chat branches (r1 #263, r2 #262, r3 #264 - 32 research scripts/modules, cross-worktree imports resolved to local ones), so this is the single living research archive going forward. The findings corpus lives outside git: set `$WMO_ROUTING_DATA` (or symlink `.wmo/routing-data`) to the shared data root to reproduce anything here. Dashboard and all six figures re-render bit-identical to the pre-rebuild tip; the champion promotion gate re-ran PASS (+0.0104, 5/5 seeds, -26.6%).


SilenNaihin:
Lineage note: the canonical end-to-end result PR #311 builds on this study and is now the program's primary research PR — this PR remains the method-selection record it references. (No content moved; see #311's header.)

PR #268, opened 2026-07-26
Off-policy distillation: a first-class mode for continual training from collected experiences

kfallah:
First cut of the launch flow: **use Tinker to continually train a student from collected experiences**, off-policy.

Off-policy here means training on a **teacher's own trajectories** (teacher generates, student learns the tokens), as opposed to the on-policy path in #257 where the student samples and the teacher scores those tokens via `compute_logprobs`.

## The useful discovery

Off-policy distillation is **already in the tree**, unnamed. `_DistillRun._warmup()` (`wmh/distill/loop.py:2015-2143`) is a complete off-policy hard-target CE phase today: `_collect_warmup_trials` runs the teacher provider over every train task, the keep filter drops failures, `build_datums` merges spans into token-exact datums, then `forward_backward(datums, CROSS_ENTROPY_LOSS)` + `optim_step`. `build_datums` is provenance-agnostic - `TrainDatum._check_alignment` enforces only token/mask alignment, never who sampled.

So this lane is mostly **promotion, not construction**. What `_warmup` is not: it is a one-shot pre-phase doing full-batch passes over the entire datum list (`loop.py:2024-2029`), gated inside an on-policy run whose `TrainConfig.steps` is `ge=1`. You cannot express "zero on-policy steps, N epochs off-policy" as a configuration.

## Teacher choice: same-family, not GLM-5.2 cross-tokenizer

`tokenizer_fingerprint_check` (`wmh/distill/teacher.py:542-598`) hard-raises on the token-exact path, and its own error text says to pick a same-family teacher sharing the student's tokenizer. Cross-tokenizer teaching needs a generation-side byte-alignment path; what exists on `feat/distill-xtoken` is scoring-side. So the first cut is a large Qwen teaching Qwen3.5-9B; GLM-5.2 x-token stays a follow-on.

## Plan

1. **This PR** - `[offpolicy]` config section + a minibatch/epoch executor, with `_warmup` collapsed onto it so every existing warmup test passes unchanged at `minibatch_datums=0`.
2. Harbor-free teacher regeneration + an append-only experience corpus.
3. Corpus-NLL holdout gate, and a whole off-policy run expressible without harbor (`wmh distill run`).
4. Serving: tool pass-through, pool-entry plumbing, so the trained student is actually reachable.
5. Opt-in request capture on `wmh serve` - the only training-grade source of "new traces".
6. `POST /v1/policy/reload` so "continually" does not mean "restart the server".
7. `wmh distill watch` - the continual driver, one gated generation per round.
8. Self-ROUGE-L collapse hard stop (borrowed as pure functions from openclaas `claas/eval/metrics/collapse.py`, no dependency).

## openclaas

Evaluated and **not adopted**. It trains self-distillation (frozen base re-prompted with hindsight), not teacher-to-student, and its `/v1/feedback` requires a hit in its own `CompletionCache`. Worth borrowing two small things rewritten in wmh idiom: `rouge_l_score` / `measure_self_rouge_l` (the one collapse detector wmh lacks) and the shape of its sampler-refresh route.

Stacked on #257. Draft until the first implementation lands.


kfallah:
## PR 1 landed on the branch

`[offpolicy]` config + a minibatch/epoch executor, with `_warmup` collapsed onto it.

**16 files, +2,361/-130.** `3213 passed, 18 skipped` (baseline 3167, +46 new tests, zero regressions). `ruff check`, `ruff format --check`, `ty check` all clean.

### Warmup collapse is bit-identical
`_warmup()` is now `run_offpolicy(..., minibatch_datums=0, shuffle_seed=None, checkpoint_every=0)`: same single `forward_backward` over the whole datum list plus one `optim_step` per pass, same order, same `train_tokens` charge, same `WarmupMetrics` rows, same `WarmupRecord`, same forced sampler refresh. The 404-line warmup section of `loop_test.py` was verified **byte-identical** to `3a65893`.

### Resumability
`offpolicy-cursor.json` is written at a `checkpoint_every` cadence, always *after* the `save_state` it names, carrying `steps_completed`, the next `(epoch, minibatch)`, corpus size and the `tinker://` state. `write_offpolicy` deletes it, so record and cursor are never both present. A cursor whose recorded corpus size no longer matches raises an actionable error rather than silently retraining a prefix the student already carries.

`offpolicy.py` is deliberately store-free: `plan_offpolicy_steps` is a pure expansion of `(corpus size, schedule)` into steps, which is what makes "resume" equal "skip the first N positions".

`[warmup]` coexists (its tests pin the surface) but a validator now rejects enabling both, since they train the identical objective on the identical corpus.

### Falsification
Two passes, source stashed and restored byte-identically. The stronger pass reverted **only** `loop.py`, leaving config/executor/store in place:
- resume test -> `Failed: DID NOT RAISE RuntimeError` (the phase never ran, so there was nothing to interrupt)
- minibatch test -> `AssertionError: At index 0 diff: 'importance_sampling' != 'cross_entropy'` (the 4 CE steps are missing)

Also driven end to end through the real CLI cost path, and the run-dir artifacts read back directly: 12 `phase: "offpolicy"` rows across two sessions with no duplicates, `offpolicy.json` with `steps: 12`, cursor cleared, sampler tagged `offpolicy-step-0000`.

### Open question for review
`train.steps` is still `ge=1`, so **a pure off-policy run with zero on-policy steps is not yet expressible.** The phase runs before the OPD loop and the example config keeps `train.steps` small so the gate still reads a student that has seen its own states. If the first cut should be off-policy *only*, relaxing that bound is the next change.

Minor: both CE phases still read and write `warmup-trials.json`. Reusing it lets an off-policy run and a warmup run share a collection and avoids duplicate store code, but the legacy name reads oddly in an off-policy-only run dir. Cheap to rename with a back-compat read.

SilenNaihin:
Coordination from the training lane (tau2 distill), so these two do not collide or duplicate.

**Relationship: #309 is your text-input front end, not a competing executor.** [#309](https://github.com/experientiallabs/world-model-optimizer/pull/309) adds `wmo/distill/text_episodes.py`: teacher episode TEXT from any provider (Fireworks, OpenRouter, Azure, a recorded corpus) re-encoded under the STUDENT's renderer into `TrialRecord`s with synthetic spans, emitted as a `WarmupTrialsManifest`. It deliberately adds no executor and no scheduling: today it reaches training through `warmup.trajectories_from`, and when your `[offpolicy]` executor lands it should feed that instead, unchanged, because it produces exactly the `TrialRecord`s `build_datums` already consumes. If you would rather own the manifest-emitting entry point, say so and I will adapt to your seam; I have no attachment to the surface, only to the re-encoding being correct.

**Base status, since it affects your rebase:** `feat/distill-capability` is not an ancestor of main, but its content IS in main via #257's squash merge (merged 2026-07-26T17:41Z). So the retarget is a rebase of `feat/distill-offpolicy` onto main rather than a base-branch change. I have not touched your branch; it is yours to rebase.

**One shared-code change of mine you will need to absorb, and it is small.** In `wmo/distill/data.py` I added a safety invariant that matters directly to off-policy work:

- `TokenSpan.logprobs_are_placeholders: bool = False` (in `wmo/providers/tinker.py`)
- `TrainDatum.hard_targets_only: bool = False`, set in `_merge_trial_spans` when any contributing span carries the flag
- `attach_advantages` raises on any datum with `hard_targets_only`

Why: re-encoded text has no sampler logprobs, and `importance_sampling`/`ppo` derive `teacher_lp - sampled_lp` FROM them, so text-derived datums on an advantage loss would train against a fabricated baseline. Cross-entropy is unaffected (`to_tinker_sft_datums` sends `target_tokens` and `weights` only). All defaults are `False`, so every existing sink, manifest, and test loads unchanged.

**Two measured findings from cycle 1 that your executor will meet:**

1. **Text-derived episodes are one datum per TURN, not per episode, and that is correct.** A Qwen3.5 generation prompt ends `<|im_start|>assistant\n<think>\n` (thinking primed) while the same turn rendered as history has no think block, so the token streams diverge by exactly those two tokens. Chaining them into one datum would buy a train/serve mismatch. Expect a ~92% fragmentation rate on the text path specifically; on the sampled path the metric still means what it always meant.
2. **A warmup-only run (`train.steps = 0`) currently has zero degeneration-tripwire coverage,** because tripwires are computed per training step. If your epoch executor is the new home for supervised passes, that is the natural place to fix it. Related: warmup inherits the OPD default `learning_rate` (1e-4) where the anchor config deliberately uses 5e-5 for warmup.

Cycle 1 context in one line: warmup-only from Qwen3.6-27B into Qwen3.5-9B on real tau2 gated REJECTED, diagnosed as teacher headroom (73.3% teacher vs 71.7% student, so nothing to distill), not a training defect. Full note and per-task rows are in DECISIONS.md.

PR #273, opened 2026-07-26
Rename world-model-harness to world-model-optimizer (wmh -> wmo)

kfallah:
The GitHub repo is already renamed. This brings the code, docs and packaging in line.

## What changed

Total rename, no compatibility surface — no aliases, no import shim, no re-publish of the old
PyPI name. Version stays **0.1.0**.

| Surface | Before | After |
|---|---|---|
| Python import package | `wmh/` | `wmo/` |
| Console script | `wmh` | `wmo` |
| PyPI distribution | `world-model-harness` | `world-model-optimizer` |
| State directory | `.wmh/` | `.wmo/` |
| Environment variables | `WMH_*` | `WMO_*` (all 46) |
| OTel span attributes | `wmh.*` | `wmo.*` |
| HuggingFace datasets | `wmh-<benchmark>-traces` | `wmo-<benchmark>-traces` |
| E2B template alias | `wmh-hb-v1-` | `wmo-hb-v1-` |

Plus W&B projects, PostHog event names, renderer registry keys, wire headers, class names,
filenames, docs and every command example.

**Zero occurrences of `wmh` / `WMH` / `Wmh` remain anywhere in the tree.**

648 files: 493 renames, the rest content.

## Two fixes riding along

- **Python↔TypeScript wire-key guard.** The field the pi runner reads tool-call parse errors off
  is a bare string shared across a process boundary — a one-sided edit reads `undefined`, empties
  the error list, and misclassifies every unparsed tool call as `no_tool_call`, with nothing in
  lint, types or the suite to catch it. Now pinned in the existing TS/Python drift test next to
  `MAX_NONACTION_TURNS`. Verified the assertion fails when the pair is split.
- **The web gate had rotted.** Its trigger watched `examples/**/models/**`, which no longer exists;
  `build-index.mjs` actually walks `packages/environment-capture/*/models/`. So the job had not run
  since 2026-07-11 and `src/data/index.json` went stale on main (independently of this PR —
  reproduced on a pristine `main` checkout). Trigger repointed and the index regenerated here.

## Callouts

- Run configs living outside this repo (training box, existing run dirs) that name the old
  renderer keys or `WMH_*` variables need updating. These fail loudly, not silently.
- Existing `.wmh/` state directories are not migrated — the new state dir is `.wmo/`.
- Deployed secrets (Vercel / Modal / E2B / CI) need their `WMH_*` keys renamed to `WMO_*`.
- Existing harbor job dirs are not resumable across this change; delete rather than resume.
- `platform/` is untouched. It pins an old git rev and GitHub redirects the renamed URL.

## Verification

- `just gate` green: ruff, ruff format, `ty`, **3282 passed / 18 skipped**
- Test parity: 225 test files before and after, 3300 collected — `testpaths` did not silently stop collecting
- Wheel built and installed exactly as CI does: `import llm_waterfall, wmo` and `wmo --help` both OK
- `importlib.metadata.version("world-model-optimizer")` resolves to `0.1.0`
- `vendor_pi.sh --check`: vendored tree byte-identical to upstream
- `web/`: eslint + tsc clean, index regeneration deterministic
- `assets/demo.gif` and `web/package-lock.json` untouched — both contain byte sequences that a
  case-insensitive replace would have corrupted

## Before releasing

Create a PyPI **pending trusted publisher** for project `world-model-optimizer`
(owner `experientiallabs`, repo `world-model-optimizer`, workflow `python-package.yml`,
environment `pypi`). PyPI does not follow GitHub repo renames, and this only fails at release time.


kfallah:
## How to review this without reading 648 files

A mechanical rename collapses to a single claim: **this diff is exactly the substitution, and
nothing else.** Verify that claim once, and the only thing left to read by hand is the residue.

### Step 1 — reproduce the mechanical part

Replay the substitution on a pristine `main` and diff the result against this branch. Everything
that matches is, by construction, not worth reading.

```bash
git worktree add -q --detach /tmp/base origin/main && cd /tmp/base
# case-SENSITIVE, exactly these, over `git grep -I -l` (text files only):
#   world-model-harness -> world-model-optimizer      Wmh -> Wmo
#   World Model Harness -> World Model Optimizer      WMH -> WMO
#   world[- ]model[- ]harness -> ...optimizer         wmh -> wmo
# excluding: assets/demo.gif, uv.lock, web/package-lock.json, wmo/harness/vendor/pi-agent/**
git mv wmh wmo
#   + examples/ingest/*_to_wmh.sh          -> *_to_wmo.sh          (6)
#   + packages/environment-capture/*/convert_to_wmh.py -> convert_to_wmo.py (4)
#   + VENDOR.md path strings, then re-hash its one line in manifest.sha256
diff -r -q --exclude=.git /tmp/base <this-branch>
```

**Result: 4 files differ. That is the entire human-review surface.**

### Step 2 — read the residue (11 hand-written lines)

| File | Lines | What |
|---|---|---|
| `.github/workflows/web.yml` | 2 | trigger repointed at `packages/environment-capture/*/models/**` |
| `wmo/harness/pi_e2b_test.py` | 6 | Python↔TS wire-key drift guard + its import |
| `uv.lock` | 3 | `uv lock` output — the dist name, 3 occurrences, no version drift |
| `web/src/data/index.json` | generated | `build-index.mjs` output; **2287 changed JSON lines, 0 mentioning `wmh`/`wmo`** — i.e. entirely the pre-existing staleness, nothing to do with the rename |

### Step 3 — check the invariants instead of the lines

These are the properties that would actually be violated by a bad rename:

```bash
git grep -c -E '(wmh|WMH|Wmh)' -- . ':!uv.lock'        # -> zero matches
git ls-files | grep -ci wmh                             # -> 0 paths
git grep -I -oh -E 'WMO_[A-Z0-9_]+' | sort -u | wc -l   # -> 46 (was 46 WMH_)
git ls-tree -r --name-only origin/main | grep -cE '^(wmh|packages)/.*_test\.py$'   # 225
git ls-files | grep -cE '^(wmo|packages)/.*_test\.py$'                            # 225 -- no tests dropped
uv run pytest -q --collect-only | tail -1                # 3300 collected
bash wmo/harness/vendor/vendor_pi.sh --check             # vendored tree == upstream
git diff --stat origin/main...HEAD -- assets/demo.gif web/package-lock.json       # empty
```

The two `--stat`-empty files matter: both contain byte sequences a *case-insensitive* replace
would have corrupted (`demo.gif` has `wmh` inside LZW image data; `package-lock.json` has `WmH` /
`wMh` / `wMH` inside base64 sha512 integrity hashes). The replace was case-sensitive on exactly
three literals, and both files are untouched.

### What the green gate does and does not prove

Proves: imports resolve, types check, 3282 tests pass, the wheel builds and `wmo --help` runs.

Does **not** prove — and so was checked separately:
- `.agents/` and `examples/` are excluded from both ruff and ty, and `testpaths` never imports
  them. Checked with `git grep -E '^\s*(from|import) wmh'` over the whole tree -> zero.
- Strings coupled across a process boundary. One real defect was found here and fixed: the pi
  runner's TypeScript read `choice.wmh_unparsed_tool_calls` while Python emitted the renamed key.
  A one-sided edit is silent — the property reads `undefined` and every unparsed tool call is
  misclassified. That pair is now pinned by an assertion, verified to fail when split.
- `testpaths` still collecting. A stale value makes pytest collect nothing and exit green; hence
  the 225-test-file and 3300-collected parity checks above.

PR #296, opened 2026-07-27
Add 'just setup': the repo's first-command entry point

SilenNaihin:
## Why

Silen's directive (joint-tau track, 2026-07-27): the repo needs one canonical command that sets everything up and creates the env, documented as step 0 of the pipeline. No such entry point existed; the justfile is the repo's designated task runner ("tasks in the justfile", repo_layout_test.py), so the command lands there rather than as a new top-level script.

## What

- `just setup`: idempotent first-run recipe. Creates `.env` from `.env.example` (never touches an existing `.env`), runs `uv sync --extra dev`, prints the next step (`wmo providers set`).
- `.env.example` gains the three keys that already have live consumers, one consumer-naming comment each: `OPENROUTER_API_KEY` (the OpenRouter provider, #284), `TINKER_API_KEY` (distillation + serving trained adapters, #250/#272), `E2B_API_KEY` (e2b rollout/eval backends).
- README Development block leads with `just setup`.

## Justification ledger

| Addition | Named consumer |
|---|---|
| `setup` recipe | the docs/cookbook step 0 (joint-tau canonical example, PR to follow) and every fresh clone |
| `.env.example` keys | wmo/providers/openrouter.py, wmo/providers/tinker.py + `wmo optimize route student` entries, the e2b backend paths |
| README line | same |

Deliberately NOT added: `WMO_COMPRESSOR_URL`/`WMO_COMPRESSOR_API_KEY` (no consumer on main yet; they ride with the compaction chat's H100 compressor-endpoint PR).

## Verification

repo_layout_test (11 passed; root-file allowlist already carries justfile/.env.example), ruff clean. The recipe body was executed directly line by line: create-from-template works, all three keys present, second run leaves .env alone. The `just` binary itself is not installed on this machine, so the recipe was not run through just; it copies the syntax patterns of the existing recipes verbatim.

Merges only after coordinator /ready-for-merge (AGENTS rule 13) + Silen eyeball.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -13,6 +13,15 @@ AZURE_OPENAI_ENDPOINT=
 # OpenAI direct (GPT 5.5)
 OPENAI_API_KEY=
 
+# OpenRouter (serves pool candidates like GLM-5.2 / Kimi, priced from its own catalog)
+OPENROUTER_API_KEY=
+
+# Tinker (distillation: student sampling, training, and serving trained adapters)
+TINKER_API_KEY=
+
+# E2B sandboxes (distillation rollouts and optimizer/eval runs with the e2b backend)
+E2B_API_KEY=
+
 # Key for OpenAI-COMPATIBLE custom endpoints (Azure Foundry resources, self-hosted vLLM).
 # Used whenever an `openai`-kind provider has an explicit `endpoint` — the real OPENAI_API_KEY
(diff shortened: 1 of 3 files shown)


SilenNaihin:
P2 addressed in e26bc919: the Development section now names just with brew/cargo install pointers.

PR #299, opened 2026-07-27
Reproduce a Tinker-trained LoRA off Tinker, on local vLLM

kfallah:
## What

Tooling to take the on-policy-distilled Qwen3.5-9B LoRA off Tinker and measure it on
TerminalBench-2 through harbor/terminus-2 against a self-hosted vLLM — checking the trained
policy against a second inference stack rather than only the one it was trained on.

**Request:** eval the Qwen3.5-9B LoRA (Tinker sampler `pi-step-0004-b1604882`, run
`qwen-train-v3`, teacher Qwen3.6-27B) on the 17-task TB2 holdout with and without the adapter,
served in vLLM on 2×H100.

**Criteria:** reproduce the published Tinker-side result — student-before 21.6% (11/51),
student-after 27.5% (14/51), paired delta +0.059, 95% CI [+0.000, +0.157].

**Machine:** `azureuser@4.154.170.26` (h100-dev-box), 2× H100 NVL 95GB.

Numbers and the run journal live in Notion, not here — this PR is tooling only.

## The adapter is not vLLM-LoRA-servable

Tinker trains with `train_unembed=True`, so the export carries `lm_head` LoRA. vLLM only
admits LoRA on embedding modules when the model class declares a non-empty `embedding_modules`,
and `Qwen3_5ForConditionalGeneration` declares `{}`. It also stores split
`linear_attn.in_proj_{q,k,v}` against the base's fused `in_proj_qkv`. Merge-then-serve handles
both.

This contradicts the current lane page and HF model card, which both state the published
adapter is loadable with `--enable-lora`. Correction posted to the lane.

## Merge verification is split in two, deliberately

`verify_merge.py` asserts every targeted module actually moved, applying the
`unembed → lm_head` rename and the split→fused qkv concat. Necessary but weak.

`merge_fidelity.py` answers the quantitative question. For a 4-step LoRA whose delta is
~0.1–0.6% of the weight norm — near bf16's ~0.4% relative resolution — most individual elements
legitimately round straight back to base. On this checkpoint only **39.97%** of adapted elements
changed, which would fail a sibling lane's ≥40% gate, yet the **delta norm ratio is 0.962**: the
update landed essentially intact and bf16 merely redistributed it. The element fraction is the
misleading statistic; the norm ratio is the honest one.

Reported alongside: the cosine between intended and realized delta (0.70–0.97 here), which says
bf16 rounding noise is comparable in scale to a few-step update — so a merged model is not
bit-equivalent to runtime LoRA application. Worth stating in any writeup.

## Environment failures recorded because they cost real time

- flashinfer's sampling kernels JIT at startup and need **both** `ninja` and the CUDA **dev**
  headers. A box can have a working CUDA runtime and still lack `curand.h`.
- Serving with `max_model_len` equal to the agent's `context_budget + max_tokens` leaves no room
  for terminus-2's proactive summarization, which sends slightly more than the budget and gets a
  400. litellm swallows the body, so the symptom is an empty `Error in proactively summarizing:`
  — trials still complete, `exception_info` stays null, and the run continues at full speed with
  compaction silently disabled. This is a trap specific to reproducing Tinker runs, whose budgets
  are tuned to sit just under Tinker's 65,530 ceiling.

## Verification

Everything here was exercised end-to-end on the box, not just written:

- merge verified: 11/11 adapter target classes changed, `mtp` and vision tower correctly
  untouched, 0 all-zero `lora_B`, norm ratio 0.962
- behavioural tripwire: 5/5 prompts produce different greedy output between the arms
- `verify_merge.py` caught its own naming bug on first run (the VL wrapper nests the text stack
  under `language_model`) and `merge_fidelity.py` caught an fp32 accumulation error that reported
  a cosine of 1.33; both fixed and re-run
- `.agents/` is `extend-exclude`d from ruff/ty per `pyproject.toml`, so the print-based CLI
  scripts do not gate

### DIFF
--- a/.agents/distill/tb2-vllm-repro/README.md
+++ b/.agents/distill/tb2-vllm-repro/README.md
@@ -1,0 +1,87 @@
+# Reproducing a Tinker-trained LoRA off Tinker, on local vLLM
+
+Takes an on-policy-distilled LoRA checkpoint produced on Tinker, merges it into a full
+HuggingFace model, serves base and merged side by side in vLLM, and evaluates both on the
+TerminalBench-2 holdout through harbor's `terminus-2` agent with E2B sandboxes.
+
+The point is to answer a question Tinker itself cannot: **does the trained policy survive
+leaving the training service?** Tinker's sampler and our vLLM are different inference stacks,
+so a lift that only exists on Tinker is not a property of the weights.
+
+Run of record: student `Qwen/Qwen3.5-9B`, teacher `Qwen/Qwen3.6-27B`, LoRA rank 32,
+sampler `pi-step-0004-b1604882` from `qwen-train-v3`. Numbers live in Notion, not here.
+
+## Two findings worth not rediscovering
+
+**vLLM cannot serve this adapter as a LoRA — you must merge it.** Tinker trains with
+`train_unembed=True`, so the adapter carries `unembed_tokens.lora_A/B`, which the export
+remaps to `lm_head`. vLLM only admits LoRA on embedding modules when the model class declares
+a non-empty `embedding_modules`, and `Qwen3_5ForConditionalGeneration` declares `{}` — so
+`LoRAModel.from_local_checkpoint` rejects the adapter outright. Independently, the adapter
+stores split `linear_attn.in_proj_q/k/v` while the base and vLLM use a fused `in_proj_qkv`,
+which is not exactly representable as a rank-32 LoRA on the fused tensor. Merging sidesteps
+both: `merge_strategy="auto"` concatenates the split deltas into the fused weight.
+
+**A merge can silently no-op, so assert coverage rather than eyeballing the output.**
+`verify_merge.py` checks that the set of base tensors that actually changed equals the set the
+adapter targets, after applying the split→fused and unembed→lm_head remaps. Sampling a few
+tensor names by hand is not enough: an obvious pick like `layers.0.*` hits `input_layernorm`,
+which `all-linear` never targets and which is *correctly* unchanged. A green "weights differ"
+from the wrong tensor is worse than no check.
+
+## Sequence
+
+```bash
+# 1. Export + merge  (CPU-only, ~19 GB out; output_path must not exist)
+python export_and_merge.py \
+  --tinker-path 'tinker://<uuid>:train:0/sampler_weights/<name>' \
+  --base-model  /scratch/hf_cache/hub/models--Qwen--Qwen3.5-9B/snapshots/<sha> \
+  --output      /scratch/repro-tb2/qwen35-9b-distill-v3
+
+# 2. Prove the merge landed before spending sandbox budget
+python verify_merge.py \
+  --base /scratch/hf_cache/hub/models--Qwen--Qwen3.5-9B/snapshots/<sha> \
+  --merged /scratch/repro-tb2/qwen35-9b-distill-v3 \
+  --adapter /scratch/repro-tb2/adapter-raw
+
+# 3. Serve both arms, one model per GPU (TP=1 each so the arms run concurrently)
+./serve_arms.sh
+
+# 4. Confirm the arms are not the same model
+python tripwire.py            # greedy decode, must differ on every prompt
+
+# 5. Evaluate  (arm-base.yaml is the canonical config; derive the other arm from it)
+harbor run -c arm-base.yaml --yes
+harbor run -c arm-distill.yaml --yes
+
+# 6. Score, paired over tasks
+python score.py --base-jobs <dir> --distill-jobs <dir>
+```
+
+## Environment notes
+
+These bit a real run; they are cheap to pre-empt.
+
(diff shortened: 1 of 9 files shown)


kfallah:
## Result: the reproduction succeeded, including the part where it isn't significant

Both arms ran the full **17 × 8 = 136 trials with zero ungraded trials and zero cancellations**, so the errored-dropped and errored-scored-zero conventions coincide exactly and no trial left a denominator.

| measure | this run (vLLM, merged bf16, k=8) | original (Tinker, runtime LoRA, k=3) |
|---|---|---|
| base / student-before | **18.4%** (25/136) | 21.6% (11/51) |
| distill / student-after | **25.0%** (34/136) | 27.5% (14/51) |
| paired per-task delta | **+0.066** | +0.059 |
| 95% bootstrap CI over tasks | **[−0.066, +0.213]** includes zero | [+0.000, +0.157] includes zero |
| graded `ctrf` base → distill | **44.9% → 40.6%** | 44.5% → 35.3% |
| tasks moved / pinned | 5 / 12 of 17 | 2 / 14 of 17 |
| gate at 0.7 × teacher = 34.3% | **NOT MET** | NOT MET |

Three things reproduced independently: the point estimate, the graded-down-while-binary-up inversion, and both of the original's only two movers going to 1.000 (`merge-diff-arc-agi-task` 0.250 → 1.000, `constraints-scheduling` 0.375 → 1.000).

**k=8 bought per-task resolution but no significance**, exactly as predicted — the paired SE is dominated by between-task variance and floors at 0.176/√17 = 0.043 regardless of k. What the extra attempts did buy was exposing two regressions k=3 was too coarse to see (`nginx-request-logging` −0.250, `pytorch-model-cli` −0.500), which cancel much of the gain.

### Two caveats that belong with the number

**55.1% of distill trials errored against base's 14.0%**, almost entirely the 2,700s episode wall — the distilled model's median episode ran the full wall (p50 2,700s vs base's 1,347s) because its turns take ~2× longer to decode on one H100. Scored zero inside the denominator, so the arithmetic is sound, but +6.6pp was produced with over half its episodes truncated. It's a floor. A longer-wall arm is the obvious follow-up and hasn't been run.

**The efficiency result is better measured than the solve rate.** Base consumed 211.2M input tokens to distill's 107.6M for comparable output, with turns/episode falling 66.1 → 33.4 (original: 53.6 → 29.4). Given ~94% of this lane's cost is context re-reading, halving it is a tighter and more useful effect than the solve-rate delta.

### Tooling verification

Every script here was exercised end-to-end on the box. `verify_merge.py` caught its own naming bug on first run (the VL wrapper nests the text stack under `language_model`); `merge_fidelity.py` caught an fp32 accumulation error reporting a cosine of 1.33, fixed with float64. Merge proof: 11/11 adapter target classes changed, delta norm ratio **0.962**, 0 all-zero `lora_B`, and 5/5 greedy prompts differing from base.

Full writeup on the Notion lane page. Numbers and per-run configs stay there, not in this PR.

kfallah:
## Correction: the "floor" framing above is withdrawn

The comment above says the 55.1% distill error rate makes +6.6pp "a floor." That overstated it. Classifying every exception and crossing it against per-task solve rates:

| | timeouts | on tasks the arm **never** solves | on **winnable** tasks |
|---|---|---|---|
| base | 19 | 18 | **1** |
| distill | 75 | **68 (91%)** | **7** |

Every error in both arms is one mechanism — `AgentTimeoutError` at the 2,700s wall, 19/19 and 75/75, no mixture. But 68 of distill's 75 timeouts land on the 12 tasks scoring 0.000 in both arms: `build-pov-ray` 8/8, `dna-assembly` 8/8, `winning-avg-corewars` 8/8, `write-compressor` 8/8, and so on. **The truncation happens where there was nothing to win.**

Conditioning on winnability, distill lost at most 7 trials to the wall and base 1. If every truncated winnable trial would have succeeded — optimistic — the paired delta moves +0.066 → **~+0.11**. So the handicap is worth about +0.04, not "half the run."

The underlying observation still holds and still deserves the longer-wall arm: distill's median episode really does run the full wall (p50 2,700s vs base's 1,347s) because its turns take ~2× longer to decode on one H100. It just isn't what produced the delta.

### One mover is speed, not capability

Base timed out on `distribution-search` **8/8** and scores 0.000; distill timed out 4/8 and scores 0.500. That task's +0.500 comes substantially from finishing inside the budget rather than from solving better. Decomposed:

- `constraints-scheduling` +0.625 — **zero timeouts in either arm**, pure capability
- `merge-diff-arc-agi-task` +0.750 — one base timeout, essentially capability
- `distribution-search` +0.500 — largely speed-mediated

Two of three are clean. Finishing in budget is a real advantage for a deployed agent but it is a different claim from solving more, and the original writeup did not separate them either. A longer-wall arm would.

**Generalizable lesson, added to the README:** never quote a raw error or attrition rate as a bias magnitude without conditioning on whether the lost trials were winnable. 55% sounds like it dominates the result; it was worth ~+0.04.

PR #304, opened 2026-07-27
docs/cookbook: the canonical end-to-end example (tau-bench) + usage map

SilenNaihin:
## What this is

The canonical cookbook: `docs/cookbook/tau-bench.md` walks the whole pipeline once on one
benchmark, in pipeline order, where every step is **one real CLI command plus the artifact it
creates**. Plus `docs/usage.md`, the terse reference-style map of the CLI surface (purpose and
artifact per command) that keeps every other doc from re-explaining what a command is.

Steps covered: `just setup` -> `wmo build` -> `wmo providers set` -> `wmo optimize model` (staged,
with a stage-by-stage subsection so a reader can drop to `route sweep`/`fit`/`tune`/`report`) ->
optional `wmo optimize distill run` + `route student` -> optional compression arms -> `wmo serve`
with its savings and dial routes.

## Justification ledger

- **Named consumer**: Silen's cookbook directive (an end-to-end example where each pipeline step
  is one real command plus its artifact, told on tau-bench).
- **Second consumer**: the canonical benchmark writeup, which will link here for the protocol
  instead of restating the commands.
- **`docs/usage.md`**: the cookbook is narrative and one-path by design, so it cannot double as
  the map of the surface. One page carries that instead of a "see also" tail on every reference.
- Both rows are in `docs/README.md`'s justification table, as rule 5 requires.

## Commands are verified, not invented

Every command was checked against this branch's CLI with `--help`; no flag here is guessed. The
transcripts in the doc are **real output**, produced at $0 on the free paths:

- `wmo build --fidelity low --embed-provider hashing` on the real 1033-trace tau corpus: the split
  line, the ready card, and the artifact tree are that run's.
- `wmo providers set --pool-model ...`: the real confirmation lines and the resulting `pool.toml`.
- `wmo optimize model --scenarios 8`: the real plan table, including the paragraph explaining that
  the projection is the candidate half only and the simulator's own calls measured 7x it.
- `wmo optimize route pin` and `wmo serve`: the real `/v1/models`, `/v1/endpoints/*/savings`, and
  `/v1/endpoints/*/config` bodies, including the empty-state `estimate_basis` and a static
  endpoint's `"dialable": false`.

The payoff block's quality/cost/latency figures are **clearly marked placeholders** ("your numbers
will differ") because this walk's sweep was not carried to completion. Nothing is fabricated.

## Provenance discipline

Stated once near the top and inherited by every number below it: world-model simulated is not
real-episode; the cost metric is cache-adjusted effective cost per completed task (and the
sweep's single-shot list-price figures are labelled as the stricter pre-cache view); savings are
estimates against a named anchor. A closing section says plainly what the walk does not prove.

## Layout plumbing

- `wmo/repo_layout_test.py`: the docs-layout allowlist gains `docs/cookbook/*.md` and the root
  `docs/usage.md`.
- `AGENTS.md` rule 5: one sentence each for `cookbook/` and `usage.md`.
- `docs/README.md`: layout bullets plus a justification row per new doc.

## Notes

- The ladder-results doc (the `world_model_findings.md` layer) lands separately, after the grid
  runs. This PR is the protocol, not the results.
- **Merges only after the joint-tau master's review.**

## Gate

`ruff check` clean, `ruff format --check` clean, `ty check` clean, `pytest -q` 4024 passed /
18 skipped. No em dashes in the diff; no `.agents` mention in `docs/**` (test-enforced).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -168,8 +168,11 @@ uv run pytest -q
    do not force it into an unrelated directory merely to preserve the current list. What each
    surface is for:
    - `docs/`: **reviewed public documentation** in `docs/research/` (completed research writeups
-     and their rendered figures under `docs/research/figures/`) and `docs/reference/` (how-to
-     references verified against main). Nothing else: raw result JSONs, vector sources, design
+     and their rendered figures under `docs/research/figures/`), `docs/reference/` (how-to
+     references verified against main), and `docs/cookbook/` (end-to-end walks through the whole
+     pipeline on one benchmark, each step one real CLI command plus the artifact it creates),
+     plus the single root page `docs/usage.md` (the terse map of the CLI surface: one line of
+     purpose and one artifact per command). Nothing else: raw result JSONs, vector sources, design
      notes, drafts, and proposals all live in `.agents/docs/`. `docs/README.md` indexes every
      doc and records its purpose. Update or remove superseded material only after checking
      references and retaining durable evidence. `docs/`
--- a/docs/README.md
+++ b/docs/README.md
@@ -10,12 +10,18 @@ here must justify its existence in the table below; a doc that can't gets delete
   `research/figures/`. Raw result JSONs, vector sources, and experiment logs are not docs.
 - **`reference/`** — how-to references for user-facing systems, verified against current `main`
   at promotion time.
+- **`cookbook/`**: end-to-end walks through the whole pipeline on one benchmark, each step one
+  real CLI command plus the artifact it creates. One file per benchmark.
+- **`usage.md`**: the only root page besides this one. The terse map of the CLI surface, one line
+  of purpose and one artifact per command.
 
 ## Why each doc exists
 
 | File | Justification |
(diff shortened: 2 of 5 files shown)


SilenNaihin:
Merge-order dependency (added after the initial review pass):

The "Running it unattended" section of `docs/cookbook/tau-bench.md` is written against **#305**'s semantics (`--dry-run`, and a non-TTY spending run without `--yes` exiting 2). Verified against #305's diff, not paraphrased. **#305 must merge before this PR**, or that section describes behavior that does not exist yet.

Deliberately scoped: #305 changes `wmo optimize model` only. `wmo optimize route sweep` still proceeds without confirmation on a non-TTY (`wmo/cli/route_app.py:603`), so the stage-by-stage section says so explicitly rather than letting a reader generalize the refusal. If that asymmetry gets closed later, the paragraph at `docs/cookbook/tau-bench.md:385` is the one to update.

PR #308, opened 2026-07-27
fix(routing): Reserve held-out scenarios for policy reports

yashwardhanchaudhuri:
## Issue

The canonical routing workflow fitted a policy and generated its improvement report from the same outcome matrix scenarios. Although those scenarios were held out from world-model prompt optimization, they were not held out from router fitting.

For kNN policies, reported queries could retrieve their own fitted evidence row while the report described the result as held-out performance.

## Implication

The report could overstate routing quality and cost improvements because it measured in-sample router performance. This affected evaluation integrity and the credibility of endpoint improvement claims, without changing runtime availability or security.

## Reproduction

The one-command workflow used the same `matrix.json` for both stages:

```text
sweep -> matrix.json -> fit
                     -> report
```

A reproduction confirmed that every reported scenario ID was also present in the router's fit set and kNN evidence bank.

## Assertions

The fix establishes these invariants:

- Router fit and report scenario IDs are disjoint.
- Assignment is deterministic and independent of matrix row order.
- Seventy percent of scenarios are used for fitting and thirty percent are reserved for reporting.
- At least one scenario remains on each side.
- Reports refuse to claim held-out performance when no untouched scenario exists.
- Completely separate evaluation matrices remain supported.
- Legacy kNN policies recover fit IDs from their evidence bank.

Validation completed with 182 focused tests, Ruff, Ty, and 4,088 repository tests passing.

## Supported fix

Persist the router's fit scenario IDs in the policy artifact. The fit commands use a deterministic scenario-hash partition, while reporting excludes all recorded fit IDs before calculating headline, candidate, cost, latency, and model-mix results.

Report artifacts now include the exact evaluated scenario IDs for auditability. Legacy rank policies without fit provenance must be refitted before producing a held-out report.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ wmo build --file traces.jsonl --name my-endpoint
 # Score every registered model on held-out tasks from your traces
 wmo optimize route sweep my-endpoint --traces traces.otel.jsonl
 
-# Turn those measurements into a routing policy
+# Deterministically reserve 30% for reporting and fit on the other 70%
 wmo optimize route fit matrix.json --kind knn \
   --out .wmo/models/my-endpoint/policy.json
 ```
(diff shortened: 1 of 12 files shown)


SilenNaihin:
Rebased onto current main and merged as #339 (c55d82c7) with full credit — thank you for catching the fit/report leakage; the deterministic split + policy-recorded fit ids design was exactly right. Two rebase seams were resolved (main had grown typer.BadParameter wrapping and the plan-table label); everything else is your patch as written.

PR #320, opened 2026-07-27
Validate `wmo eval` inputs before it runs, not after

kfallah:
## What was broken

`wmo eval` deferred every check to the point of use. The trace paths, the `--out` destination, the suite name, the failover chain, the optional `viz` extra and the closed-loop-only flags were all discovered mid-flight, or never. Six of them surfaced as raw Python tracebacks, several *after* the paid eval had already finished, and two failed silently.

One repro a reviewer can paste (no provider calls, since the empty file yields zero steps before any model is constructed):

```bash
mkdir -p /tmp/wmoeval && cd /tmp/wmoeval && : > empty.jsonl
wmo eval empty.jsonl --out /tmp/wmoeval/nodir/report.json
```

Before: the scorecard prints, then

```
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/wmoeval/nodir/report.json'
```

The whole report is discarded because `_write_ad_hoc_eval_report` calls `path.write_text()` with no parent `mkdir`, while the sibling suite and grid paths both create it.

After: the parent is created before the eval starts, so the report lands; and if `--out` names a directory it is a clean usage error instead of an `IsADirectoryError`.

## What changed

All of this is in `wmo/cli/app.py`, in the house style: `typer.BadParameter` with the exact next command to run. No blanket try/except was added at the entry point.

- **`_prepare_out_path(out)`**, called once at the top of `eval_()`, creates `--out`'s parent (as `optimize route fit --out` and the eval suite/grid paths already do) and rejects a directory. It covers the ad-hoc, suite, grid and closed-loop flows, so the bare write in `eval_closed_loop.py` is fixed too.
- **`_run_eval_files` rejects a directory**: the guard only checked `path.exists()`, which a directory satisfies; it now also checks `is_file()` and names `<dir>/traces.otel.jsonl`.
- **`_require_viz_extra()`** probes matplotlib/seaborn with `importlib.util.find_spec` and raises a usage error naming ``uv sync --extra viz`` / ``pip install 'world-model-optimizer[viz]'``. It runs *before* the work in `eval grid` (which used to spend the entire grid and only then die on `import matplotlib`), and in `eval grid-plot`, `eval grid-heatmap`, `research plot-concurrency` and `research plot-concurrency-combined`. The plotting imports stay lazy; nothing catches `ImportError`.
- **`_require_scorable_steps()`**: input that yields no OTel GenAI traces no longer prints a plausible `OVERALL fidelity=0.000 over 0 held-out steps` at exit 0. It says the input held no traces and names the next command, then exits 2. It runs on both the ad-hoc flow (where it can also suggest `--mode closed-loop` for a tasks file) and, per Greptile's P1, on `wmo eval run <suite>`, where the zero-step scorecard was additionally *saved* under `--results-root` and resurfaced in `wmo eval results` as a real 0.000 measurement. A suite whose config sets `files = []` is a usage error too.
- **`_resolve_eval_suite_or_usage()`** wraps `resolve_eval_suite` for `eval run <suite>` and `eval grid <suite>`, matching the tested `_resolve_example` contract ("usage error, not a ValueError traceback"). The well-written message and its `available:` list are preserved.
- **`--chain`**: `provider_or_chain`'s `ValueError`/`FileNotFoundError` (missing `.wmo/fallback.toml`, unknown chain, multi-chain file with no `default`) become usage errors two lines after the guard that already did this for a missing trace file.
- **Closed-loop-only flags are rejected in open-loop.** `--name/--root/--k/--max-turns/--harness/--harness-backend/--eval-concurrency/--e2b-template` were accepted and silently discarded, so the README's closed-loop command minus `--mode closed-loop` quietly ran a different paid evaluation. Uses the same `_explicit(ctx, param)` guard shape as `wmo optimize harness`'s world-model/harbor flag split; `--e2b-template` supplied via `$WMO_E2B_TEMPLATE` does not trip it. `--threshold` is likewise scoped to `wmo eval agreement`.
- **Help text**: the `tokens` argument and the Flows list now name `grid`, `grid-plot`, `grid-heatmap` and `agreement` (six options were documented as belonging to `wmo eval grid`, a token the help said did not exist); `wmo eval list` and `--examples-root` now point at `packages/environment-capture/<task>/evals/` where the suites actually live; the empty-args error lists every flow.

## Tests

Thirteen new tests in `wmo/cli/app_test.py`, one per behaviour, each pinning exit code 2 rather than a traceback (the two suite ones also assert that nothing was persisted). `uv run pytest` is green (4161 passed, 20 skipped), as are `ruff check .`, `ruff format --check` and `ty check`.

## Audit findings closed

`serve-eval` 8, 9, 11, 41, 42, 43, 47, 48, 49 and `research` 95.

Not in this PR (separate themes): finding 45 (`[models.agent]` eaten by rich markup in the same docstring), 10 (suite roots unreachable from a pip install) and 12 (`wmo eval` ignores the configured `[models.worker]`).

### DIFF
--- a/wmo/cli/app.py
+++ b/wmo/cli/app.py
@@ -8,6 +8,7 @@ models are named (`--name`), stored under `<root>/models/<name>/`, and listed wi
 
 from __future__ import annotations
 
+import importlib.util
 import json
 import logging
 import random
@@ -41,7 +42,10 @@ import wmo.providers as providers
 from wmo.cli.agent_session import register as register_agent_session_commands
 from wmo.cli.e2b_cmds import register as register_e2b_commands
 from wmo.cli.eval_closed_loop import run_agreement, run_closed_loop
-from wmo.cli.harness_app import harness_app, optimize_app
+
+# `_explicit` is package-private to wmo.cli, not module-private: `wmo eval` rejects
+# mode-inapplicable flags exactly the way `wmo optimize harness` already does.
+from wmo.cli.harness_app import _explicit, harness_app, optimize_app
 from wmo.cli.ingest_cmd import ingest as _ingest_command
 from wmo.cli.model_roles import configured_role_configs
 from wmo.cli.platform_cmds import register as register_platform_commands
@@ -82,6 +86,7 @@ from wmo.engine.build import DEFAULT_TRAIN_SPLIT, ingest, split_traces
 from wmo.engine.build import build as run_build
 from wmo.engine.demo import run_demo
 from wmo.engine.eval_suites import (
+    EvalSuite,
     discover_eval_suites,
     list_eval_results,
     resolve_eval_suite,
@@ -158,9 +163,14 @@ _console = Console()
 _CHECK = "[green]✓[/green]"
 
 # Module-level singleton: a typer.Argument call can't be a default inline (ruff B008).
+# Every dispatched flow is listed: the `grid*` family owns six of this command's options
+# (--val-frac, --models, --gepa-prompts, --dataset-label, --limit-traces, --judge-model), so
+# leaving its tokens out of the help made those flags reference an undiscoverable subcommand.
 _EVAL_TOKENS = typer.Argument(
     None,
-    help="Trace files to score, or eval flow: list | run <suite> | results optional-suite.",
+    help="Trace files to score, or eval flow: list | run <suite> | results optional-suite | "
+    "grid <suite> | grid-plot <result.json>... | grid-heatmap <result.json>... | "
+    "agreement <a.json> <b.json>.",
 )
 # Repeatable option default hoisted out of the signature (ruff B008 forbids the call inline).
 _RESEARCH_REAL_ARG = typer.Option(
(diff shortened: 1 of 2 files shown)


kfallah:
@greptile-apps Addressed in ccd4eae5.

**P1 "Suite zero-step results persist" (`wmo/cli/app.py` L1706-1707): fixed.** Correct catch, and worse than the ad-hoc case it mirrors, because a suite result is durable: the 0.000 scorecard was written under `--results-root` and then resurfaced in `wmo eval results` as a real measurement.

`_require_scorable_steps` now runs on the named-suite path too, before the report is printed or written. It grew a keyword-only `next_step` so each flow names its own remedy (the suite path has no `--mode closed-loop` to suggest). While wiring it I also guarded a suite whose config sets `files = []`, which would otherwise have indexed an empty list building that message.

Repro, now exit 2 with nothing persisted:

```bash
mkdir -p /tmp/wmo320/examples/tiny-task/evals
printf %s\\n {\"task_id\": \"t1\"} > /tmp/wmo320/examples/tiny-task/traces.otel.jsonl
printf %s\\n files\ =\ [\"../traces.otel.jsonl\"] > /tmp/wmo320/examples/tiny-task/evals/default.toml
wmo eval run tiny-task --examples-root /tmp/wmo320/examples --results-root /tmp/wmo320/out
```

```
Invalid value: no OTel GenAI traces in .../tiny-task/traces.otel.jsonl; check the corpus
`files` in .../evals/default.toml with `wmo ingest --file .../tiny-task/traces.otel.jsonl`
```

Two tests added in `wmo/cli/app_test.py`: `test_eval_run_suite_with_no_traces_fails_instead_of_persisting_zero` (asserts exit 2 and that `results_root` stays empty) and `test_eval_run_suite_listing_no_files_is_a_usage_error`. `uv run pytest` is green (4161 passed, 20 skipped), as are `ruff check .`, `ruff format --check` and `ty check`.

Nothing else in the review needed a change: the summary section raised no other issue. I also dropped three em dashes from the added lines, which AGENTS.md forbids in new writing.

PR #323, opened 2026-07-27
Make `optimize model` and `optimize distill` own their input-boundary failures

kfallah:
## What was broken

Seven user-visible failures on `wmo optimize model` / `wmo optimize distill` / `wmo optimize
harness`, all the same root cause: a failure that is knowable at the command's input boundary
either escapes it unconverted (a Python traceback, a raw pydantic dump) or is never checked
there at all.

The worst one is the first thing a pip-installed user hits, because all three shipped reference
configs carry a `[rollout.renderers]` table:

```bash
pip install world-model-optimizer            # no extras
cp "$(python -c 'import wmo,os;print(os.path.dirname(wmo.__file__))')/distill/configs/distill-smoke-dev.toml" cfg.toml
printf '["t1"]' > tr.json ; printf '["t2"]' > ho.json
wmo optimize distill run --config cfg.toml --run-dir ./r1 --task-ids tr.json --holdout-task-ids ho.json
```

Before: a two-panel rich traceback dumping `wmo/distill/config.py`, `wmo/cli/model_app.py` and
`pydantic/main.py`, ending in `ImportError: the tinker-cookbook SDK is not installed`.
`_load_config` caught only `FileNotFoundError`/`ValueError`, and pydantic re-raises a
non-`ValueError` out of a field validator untouched.

After:

```
Invalid value: cannot load cfg.toml: the tinker-cookbook SDK is not installed; run
`uv sync --extra distill` (or `pip install 'world-model-optimizer[distill]'`) to use the
Tinker distillation provider
```

## What changed

- **`distill run` missing extra** (`wmo/cli/model_app.py`): `_load_config` converts `ImportError`
  to `typer.BadParameter`, naming the file. All four distill-extra hints (`rendering.py`,
  `data.py`, `tracking.py`, `providers/tinker.py`) gained the pip spelling beside the uv one, the
  way the harbor hints already do, since `uv sync` is not a command a pip user has.
- **`distill report` on a torn `metrics.jsonl`** (`wmo/distill/store.py`, `wmo/cli/model_app.py`):
  `read_metrics` gains `tolerate_partial_tail` (off by default, so everything that resumes a run
  still sees the damage). `report`, which advertises itself as safe on an aborted run dir, drops a
  half-written FINAL line with a note and reports the complete rows. The excuse is scoped to a
  line that fails to PARSE, which is all a torn append can leave: `append_metrics` writes one
  JSON object per line and every strict prefix of `{...}` is invalid JSON, so a last line
  `json.loads` accepts was written whole. Anything else -- a broken row above the last one, or a
  final line that parses into a non-object -- means the file lost or gained content, so it stays
  an error, now a clean usage error matching `_load_gate` / `_load_eval_report` rather than a
  traceback.
- **Pool-file failures** (`wmo/providers/pool.py`): `load_pool` wraps `TOMLDecodeError` and
  `ValidationError`. Missing, empty, malformed, and invalid-entry rosters all name the file and
  `wmo providers set`; an invalid entry is reported as `[[model]] N field: message` instead of a
  pydantic dump with an `errors.pydantic.dev` URL. This also improves `optimize route sweep`,
  `route pin` and `providers set`, which share the loader.
- **`optimize model --fallback`** (`wmo/cli/optimize_model_app.py`): validated against the
  preflighted pool at exactly `--baseline`'s boundary, so a typo can no longer render in the plan
  table as a real model, pass the spend confirmation, and fail inside `_stage_fit` after the sweep
  has been paid for.
- **tau2 preflight** (`wmo/cli/model_app.py`): the message carries the clone-and-venv commands
  instead of pointing at `packages/environment-capture/tau-bench/README.md`, which exists only in
  a source checkout (the wheel ships `wmo` and `llm_waterfall` only).
- **`optimize harness --help`** (`wmo/cli/harness_app.py`): `--backend` no longer asserts "The
  environment is always the world model", which the same command's own docstring contradicts for
  `harbor` (there `--backend` selects the task environment, docker vs E2B). `--iterations` states
  both defaults, and the world-model literal `5` is now a named constant beside
  `_DEFAULT_HARBOR_ITERATIONS` so the help and the code cannot drift.

## Audit findings closed

- `optimize distill run`: unhandled `ImportError` traceback when a config has `[rollout.renderers]`
  and the distill extra is missing (every shipped reference config)
- `optimize distill run`: `uv sync --extra distill` is the only install spelling offered, on a
  documented `pip install` path
- `optimize distill report`: traceback on a torn/partial last line of `metrics.jsonl`
- `optimize model`: pool-file failures return a raw pydantic dump with no path and no next command
- `optimize model`: `--fallback` is not validated against the pool the way `--baseline` is
- `optimize distill run`: tau2 preflight points at a source-checkout-only README
- `optimize harness --help`: `--backend` help is false for the harbor environment
- `optimize harness --help`: `--iterations` documents no default and silently differs by environment

## Tests

Every behaviour change has coverage, and each new CLI test was confirmed to fail without its fix.
`uv run pytest` (4162 passed), `uv run ruff check .`, `uv run ruff format --check`, `uv run ty
check` all green. Each repro was also driven through the real console entry point before and after.

### DIFF
--- a/wmo/cli/harness_app.py
+++ b/wmo/cli/harness_app.py
@@ -61,6 +61,9 @@ from wmo.providers.registry import get_provider
 # The default agent seed's literal CLI name: `wmo optimize harness pi harbor ...` starts from the
 # built-in pi agent and publishes new versions under the store name "pi".
 DEFAULT_SEED_AGENT = "pi"
+# The search budget each environment falls back to when --iterations is not given. They differ
+# (harbor buys more, cheaper, steps), which is exactly why --iterations' help states both.
+_DEFAULT_WORLD_MODEL_ITERATIONS = 5
 _DEFAULT_HARBOR_ITERATIONS = 10
 _HARBOR_ENVIRONMENT = "harbor"
 _HARBOR_EXTRA_HINT = (
@@ -180,7 +183,9 @@ def optimize(
         "local",
         "--backend",
         help="Where the harness PROCESS runs: local (in/from this process) or e2b (the real "
-        "pi agent inside pooled E2B sandboxes). The environment is always the world model.",
+        "pi agent inside pooled E2B sandboxes). With a world-model ENVIRONMENT that is all it "
+        "moves: the environment stays the world model. With `harbor` it also selects the task "
+        "environment (local = docker tasks, e2b = E2B tasks).",
     ),
     eval_concurrency: int | None = typer.Option(
         None,
@@ -204,7 +209,9 @@ def optimize(
     iterations: int = typer.Option(
         None,
         min=0,
-        help="Propose-and-gate steps (the search budget). 0 scores the seed only (harbor env), "
+        help=f"Propose-and-gate steps (the search budget). Default: "
+        f"{_DEFAULT_WORLD_MODEL_ITERATIONS} for a world-model environment, "
+        f"{_DEFAULT_HARBOR_ITERATIONS} for harbor. 0 scores the seed only (harbor env), "
         "the way a baseline or a frozen champion is scored on a task set.",
     ),
     proposal_batch_size: int = typer.Option(
@@ -371,9 +378,12 @@ def optimize(
         tasks_file = Prompt.ask("Task file (JSONL of task_id/instruction/gold)")
     if iterations is None:
         iterations = (
-            IntPrompt.ask("Search iterations (each = 1 delta + 1 gated eval)", default=5)
+            IntPrompt.ask(
+                "Search iterations (each = 1 delta + 1 gated eval)",
+                default=_DEFAULT_WORLD_MODEL_ITERATIONS,
+            )
             if interactive
-            else 5
+            else _DEFAULT_WORLD_MODEL_ITERATIONS
         )
 
     if backend not in ("local", "e2b"):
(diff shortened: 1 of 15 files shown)


kfallah:
> **P1 — `wmo/distill/store.py` L352-358: Complete corruption is silently dropped**
> When the final line is a complete non-object JSON value such as `[]` or `null`, tolerant mode discards it even though `append_metrics` only writes objects.

Fixed in 8349f67c. Correct, and the reasoning generalises: the excuse only ever applied to a *torn append*, and a torn append can never leave a line that parses. `append_metrics` writes `json.dumps(payload) + "\n"` where `payload` is always an object, and every strict prefix of `{...}` is invalid JSON — so a final line `json.loads` accepts was written whole, which makes it content damage, not truncation.

`tolerate_partial_tail` is now scoped to the `JSONDecodeError` path alone. The `isinstance(parsed, dict)` check raises unconditionally again, so `distill report` on a `metrics.jsonl` ending in `[]` exits 2 with `corrupt metrics row on line 2 ... expected a JSON object` instead of quietly reporting the previous step as the latest state. Both docstrings now say why the narrower rule is the whole rule rather than a conservative choice.

Coverage, each confirmed failing without the fix:
- `store_test.py::test_a_tolerant_read_still_refuses_a_last_line_that_parses`, parametrised over `[]`, `null`, `"step 1"`, `[1, 2, 3]`
- `model_app_test.py::test_report_refuses_a_last_metrics_line_that_parses_into_a_non_object`, asserting exit 2 and that the half-written-line note is *not* printed

`uv run pytest` 4167 passed / 20 skipped, `ruff check`, `ruff format --check`, `ty check` all green.

PR #345, opened 2026-07-28
Give every durable-file writer the safeguards pool.py documents (re-land of #342)

kfallah:
Re-lands #342, which was merged into its stacked base branch (`fix/pool-duplicate-model-key`)
rather than into `main`, so its changes never reached main. Same commit, cherry-picked onto current
main; the content diff is identical (23 files, 807 insertions) and the full suite passes at 4451.

## Why

`wmo/providers/pool.py` was the only writer in the package with the full set: an atomic write
through a per-call staging file, under a bounded cross-process `flock`. Its comments argue why
each safeguard matters. None of that reasoning had reached its siblings, which write the same kind
of file: a registry or a config a later command reads back, where a torn write leaves something
unloadable that nobody edited.

Three of these were reported; the sweep for siblings found the rest.

## What was broken

- **`HarnessStore.set_alias` wrote `aliases.toml` in place.** Its distill twin is
  character-identical except the noun and got the atomic fix; this copy did not even import a
  helper. A crash mid-write loses the champion pointer `resolve_version(None)` depends on, and wmo
  silently serves `latest` instead.
- **Neither `set_alias` took a lock.** Two promotions of *different* aliases each read the same
  table and the later write dropped the earlier one, both reporting success. Atomicity does not
  help: neither write is torn, the update is simply lost.
- **Fixed `.tmp` names** in `write_text_atomic` (distill), `write_json_atomic` (population), and
  the inline writers in `config/settings.py` and `config/config.py` — the exact hazard pool.py's
  comment names.
- **`OutcomeMatrix.save` was not atomic at all**, and it writes the measured output of a sweep
  that costs real money.
- **Five competing answers to "write this file durably."** `EndpointConfig.save` and the OpenRouter
  price cache hand-rolled a fixed `.partial` with no cleanup and no fsync;
  `write_artifact_atomically` was a second general-purpose atomic writer with five call sites.
  Each was strongest on a different axis, so a fix to one never reached the others.
- **`_parse_fallback_config` called `tomllib.loads` bare** while prefixing every other error with
  the path, so a malformed `.wmo/fallback.toml` gave a filename-less "Cannot overwrite a value"
  under typer's `Invalid value:`. Both `aliases.toml` readers had the same gap.

## The consolidated writer keeps the strongest of each

Nothing regresses, because every property came from one of the versions being replaced:

| property | inherited from | why |
|---|---|---|
| fsync payload + parent dir | `write_json_atomic` | without both, a power loss persists the rename and loses the data blocks |
| cleanup on `BaseException` | `write_artifact_atomically` | `OSError` misses Ctrl-C and an unencodable payload |
| staging name unique per call | `write_artifact_atomically` | uuid not pid: two *threads* share a pid |
| destination mode carried over | the in-place `write_text` | `replace` installs a new inode, so 0600 would widen to 0644 |

Measured cost of the added fsync: **0.08 ms per write, ~0.1 s across a 200-step distill run**
whose individual steps are minutes of paid rollouts.

Deliberately left alone: the mkstemp/0600 credential and session writers. They refuse symlinks and
need owner-only permissions from creation, which is a different contract, not a stricter setting
of this one.

## Tests

Six fail on the parent commit. Both concurrency tests force the interleaving with a barrier on
every read — written the obvious way the harness one **passed 8 runs in 10 unfixed**, because
`save_version` defaults to no alias, so `aliases.toml` did not exist and `aliases()` returned from
its `path.exists()` guard before reaching the patched read. It now seeds an alias first and fails
10 out of 10.

Also pinned in `wmo/core/files_test.py`: threads never publish a torn payload, an unencodable
string leaves no staging file, Ctrl-C cleans up, the mode survives, and a failure *after* the
rename reports the new contents rather than claiming the file is untouched (the parent-directory
fsync runs post-rename and some FUSE/NFS mounts reject it — the docstring says so instead of
promising otherwise).

Verified end to end: a corrupt `aliases.toml` names itself, and deleting it does fall back to the
latest version as the message claims.

## Follow-ups, not in this PR

- `HarnessStore.save_version` hardens the pointer but not the artifact: `doc.json` and the
  rendered surfaces are still plain `write_text`, and the version-dir `mkdir` races between two
  concurrent optimize runs.
- Five `tomllib.loads` sites still raise bare (`config/store.py`, `harness/source_tree.py`,
  `platform/transfer.py`, both `credentials.py`); they want one shared `load_toml_file(path, what)`.
- `HarnessStore` and `AdapterStore` are ~78% line-identical and this PR made that worse by copying
  the lock+atomic body into both; they want an `AliasTable` or a shared base.
- `tomlkit` would round-trip comments and delete `_render_sections`, `_appended_body`, the
  append/re-render branch, the legacy path, and four "your comments are gone" warnings.


kfallah:
On the P2 follow-up and the 3/5: the CLI-startup `fcntl` resolution is pre-existing on `main`, not introduced here, so I do not think it blocks this PR.

On `main`, before this branch:

```
wmo/cli/app.py:42            import wmo.cli.pool_registry
wmo/cli/pool_registry.py:45  from wmo.providers.pool import (...)
wmo/providers/pool.py:28     import fcntl
```

`wmo providers set` needs the roster lock, so the CLI has always pulled an unconditional `import fcntl` at startup through that chain. This PR relocates that import from `pool.py` to `wmo/core/locks.py`; the set of modules that reach it from the CLI is identical.

What the split does change is the direction Greptile flagged first: `fcntl` is no longer on the import path of `wmo/config/config.py`. Verified by import:

| entry point | reaches `wmo.core.locks` |
|---|---|
| `import wmo` | no |
| `import wmo.config` | no |
| `wmo/config/card.py` | no |
| `wmo/providers/pool.py` | yes (needs the lock) |
| `wmo/harness/store.py`, `wmo/distill/store.py` | yes (need the lock) |

So the net effect on platform reach is a reduction, never an increase.

Making the CLI start without `fcntl` is a real and separate piece of work: it would need the lock guarded at every use site, and `pyproject.toml` declares no OS classifiers while CI is `ubuntu-latest` only, so whether Windows is a supported target is not currently settled anywhere in the repo. Happy to file that as its own issue rather than widen this PR.

The P1 finding was correct and is fixed in ae794f61: `_fsync_directory` now logs and returns, so an exception from `write_bytes_atomic` again means the write did not land.

kfallah:
Resolving the remaining P2 thread: **Windows is not a supported target for this project**, so the CLI-startup `fcntl` resolution is not a defect to fix here.

That makes the finding unresolvable by construction rather than actionable. `wmo providers set` needs the roster write lock, so the CLI needs `flock`, so it needs `fcntl`. Removing that from import time would mean a platform-guarded or deferred import, and AGENTS.md rule 8 rules that out directly: "Do not use lazy imports for optional convenience, and do not catch `ImportError`/`ModuleNotFoundError` to silently fall back to alternate behavior." Bending that rule to satisfy a platform the project does not target would be the wrong trade.

For the record, the condition predates this branch. On `main`: `wmo/cli/app.py:42` -> `wmo/cli/pool_registry.py:45` -> `wmo/providers/pool.py:28 import fcntl`. This PR relocates that import to `wmo/core/locks.py` and leaves the CLI reach identical, while removing `fcntl` from the import path of `wmo/config/config.py`, which is the half that was a genuine regression.

The P1 finding was correct and is fixed in ae794f61.

Follow-up worth filing separately: `pyproject.toml` declares no OS classifiers and CI runs `ubuntu-latest` only, so "Windows is not a target" is a decision that currently exists nowhere in the repo. Adding `Operating System :: POSIX` classifiers would settle it in the packaging metadata and stop this being re-raised.

PR #347, opened 2026-07-28
Remove the kimi-gui-control benchmark

kfallah:
## Why

`kimi-gui-control` was captured but never published. Both candidate Hub repo ids return 401, so `wmo download kimi-gui-control` was always a dead end, and the registry entry advertised a bundle nobody could fetch. The CLI audit flagged it as a defect for exactly that reason.

## What goes

- The `CorpusSpec` in `packages/environment-capture/environment_capture/hub.py`
- `packages/environment-capture/kimi-gui-control/` — README, `convert_to_wmo.py`, `evals/default.toml`
- The benchmark's name from the list in `docs/reference/eval_grid.md`

**No trace data was ever committed.** Across all history that directory only held those three files; the payload was gitignored (`*/traces.otel.jsonl`, `*/data/`, `*/gold/`, `*/corpus/`) and the Hub push never landed. So there is no data to purge, only a registry entry pointing at nothing.

## What deliberately stays

`.agents/docs/research/benchmark-grid/grid_kimi-gui-control_{api,qwen}.json` and the surrounding README. Those record a benchmark-grid run that actually happened; retiring a benchmark is not a reason to delete the measurements taken with it. Say the word and they go too.

## Gates

`uv run pytest -q` → 4429 passed, 20 skipped. `ruff check`, `ty check` clean.

## Ordering note

PR #322 vendors the hub registry into `wmo/hub.py`, and its copy still carries this entry (plus tests asserting the registered-but-unpublished behaviour). Whichever of #322 and this lands second needs a follow-up to strip the other copy — this PR is written against `main`, where `wmo/hub.py` does not yet exist.


kfallah:
@claude resolve merge conflict

PR #444, opened 2026-08-14
Journal programmatic router calls

kfallah:
## Summary

- make `wmo.load_router(PROJECT)` reuse the exact project journal composition used by `wmo run`
- keep the public return value as the official synchronous OpenAI client with Chat Completions and Responses support
- journal ordinary unkeyed calls by default while preserving stateless Chat separation and Responses continuation lineage
- preserve standard `Idempotency-Key` replay and 409 conflicts through official SDK `extra_headers`, including across separately loaded clients
- keep load and close provider-idle, preserve loader configuration and low-level application injection, and commit buffered streams before delivery
- prove concurrent loaded clients share one project journal and select and dispatch one keyed interaction exactly once

## Exact identity

#443 is merged. This PR is retargeted directly to current `main`.

- Base: `main` at `0c8094593f34178dc46d53f6c5a81bb22c219dab`
- Head: `agent/journal-python-router-p16` at `0abf17ffba74718db9868f4cf6c0e4f03bb6f0f1`
- Exact range: `0c8094593f34178dc46d53f6c5a81bb22c219dab..0abf17ffba74718db9868f4cf6c0e4f03bb6f0f1`

The one-commit delta remains patch-identical to the reviewed P16 commit and changes only `wmo/runtime/router/application.py` and `wmo/runtime/router/application_test.py`.

## Validation

- exact restack range-diff: 1 of 1 commit patch-identical
- post-restack application/API matrix: 26 passed
- `uv run ruff check .`
- `uv run ty check`
- focused application, endpoint, journal, and run-command suite before restack: 65 passed
- concurrent loaded-client regression repeated 20 times
- exact-head full suite before restack: 721 passed, 1 skipped
- exact-head W16 router and comparison evidence: 2 passed
- source distribution and wheel build
- isolated wheel install, public import smoke, and `wmo run --help`
- independent exact-head semantic, dependency, range, and bounded-concurrency audit

### DIFF
--- a/wmo/runtime/router/application.py
+++ b/wmo/runtime/router/application.py
@@ -180,7 +180,8 @@ def load_router(
 
     Returns:
         Official OpenAI client whose Chat Completions and Responses resources call the local
-        verified router in process. Close it or use it as a context manager when finished.
+        verified router in process and durably journal every completed interaction. Close it or
+        use it as a context manager when finished.
     """
     runtime = load_project_router(
         project,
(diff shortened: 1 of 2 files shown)


kfallah:
@greptileai review

PR #451, opened 2026-08-14
Make the public terminal OTLP dataset build directly

kfallah:
## Summary

- recognize the exact owned terminal environment-capture JSONL grammar at the raw OTLP ingest boundary
- canonicalize its legacy wire IDs, timestamps, tool pairing, and conversation lineage without inventing provider or model identity
- keep generic OTLP strict and disable compatibility for any malformed line, duplicate key, mixed record, or profile drift
- document a pinned public Hugging Face dataset that works with the ordinary wmo build PROJECT TRACES command

## Direct happy path

The unmodified pinned Hugging Face file at revision 540883e451dc13d34fb50fdd36b143cb0f1fb0db now produces:

- 280 accepted traces
- 1,370 spans
- 0 normalization issues
- 280 distinct conversation lineages
- 0 model identity records
- 50 fit tasks and 20 held-out tasks

The original source SHA256 remains 21c62cba7e3372cbf03df051dc2408699fbf8ea3561ba661b599e4949f0e5d42.

## Validation

- full suite: 1004 passed, 1 skipped
- focused ingest, build, layout, and import boundaries: 63 passed
- whole Ruff format and check: passed
- whole Ty: passed
- wheel and sdist: built
- fresh Python 3.13 wheel install: imported from site-packages and loaded the exact public file as 280 traces, 1,370 spans, 0 issues
- direct positional build and exact replay: 280 accepted, 0 invalid, 50 fit, 20 held out

No merge performed.

PR #459, opened 2026-08-15
Net-negative consolidation: delete dead code, unify duplicated primitives, trim redundant tests

kfallah:
## Summary

Post-refactor cleanup pass over the whole package: net **-2,124 lines** (2,407 added, 4,531 deleted across 143 files), with the full gate green (1,240 passed, 1 skipped in 4:37) on a tree merged through main up to #457.

Every change came from a finding that was independently discovered, adversarially verified against the real call graph, implemented, and re-reviewed by two further adversarial passes on the final diff. Greptile scored the PR 5/5 with zero inline comments; the two CodeQL weak-hash alerts were dismissed as false positives (content-addressed digests and an idempotency-key privacy digest, both `usedforsecurity=False`, no password storage anywhere). The net figure has tightened across merges because main's own cleanups (#456, #458) kept landing overlapping deletions; what remains is exactly this branch's unique work.

### 1. Dead code deleted

- `wmo/runtime/models/roles.py` role-preflight machinery, the unused `LMRubricProposer`, Harbor E2B template provisioning, the dead `TraceSource`/`LocalTraceSource` abstraction on `compose_router` (removes one unused public export).
- Sandbox-ledger reaper surface, deeper than main's cleanups: also `_owner_pid`, the `owner_pid` field, and `fully_released` (zero external references).

### 2. Existing types instead of new ones

- Native Responses wire parsed with `openai.types.responses.Response` and its typed output items instead of a hand-rolled walker.
- `pi.py` delegates its OpenAI chat-completions bridge to the shared `endpoint.py`/`streaming.py` implementation, with Pi-tolerant message subclasses (`extra="ignore"`, null `tool_calls`) so OpenAI-legal wire variations from the independently versioned Pi CLI cannot abort an episode, and the `content` key survives on tool-call-only replies. Regression tests cover all three variants.
- `ProviderModelSelection` nests `ModelCapabilities`; the `--model-json` contract nests capability keys, and #457's new picker flow constructs and reads the nested shape (`model_selection` collapsed to pass `ModelCapabilities` whole).
- Ten hand-rolled timezone field-validators replaced by `pydantic.AwareDatetime`.
- `RuntimeAcceptedEvent` composes `RuntimeInteractionIdentity` + `RuntimeAcceptance`. Note: this changes the journal/snapshot record shape; no released wmo ever wrote the old shape (the journal landed after v0.3.0), so there is no legacy state to migrate.

### 3. One owner per primitive

- `canonical_jsonl_bytes` + `sha256_bytes` replace nine JSONL serializer copies and three digest copies; `envelope_matches_manifest` replaces ~21 hand-built 5-field tuple comparisons; `read_artifact_json` is the single verified typed-artifact read; `ArtifactStore.write_or_replay`/`write_or_verify_exact` replace four idempotent-write copies; `combine_economics` replaces three usage aggregators. Input canonicalization is main's `sorted_unique_inputs` everywhere, with the domain `error_type` preserved at every call site (including the three sandbox engine sites).
- Provider layer: the hand-rolled HTTP clients collapse into a shared `ProviderHttpClient` template with a table-driven `registry.resolve`; wire validators shared in `errors.py`; the dead `extra_headers` parameter removed. Azure and Bedrock (#453) keep their explicit branches and import shared retry/timeout defaults from `base.py`; #457's `listing.py` points at the merged OpenRouter home. Anthropic now records cache writes in `Usage.cache_write_input_tokens`, so pricing takes the exact three-way split instead of billing uncached input at the cache-write ceiling.
- CLI: shared `ROOT_OPTION` and a `usage_error` contextmanager replace 15 try/except copies; `preflight_automatic_router` takes `AutomaticRouterOptions` whole, and `maximum_concurrency` is bounded at preflight with its siblings.
- SFT/leases file plumbing reuses `wmo.common.core.files`; the acceptance pointer keeps its replace-the-symlink-itself guarantee behind `write_bytes_atomic(follow_symlinks=False)`, mode carryover stats the destination without following symlinks, and a regression test pins the swapped-link case.
- The judging protocol's scalar judge-response models come from common's now-public `RawJudgment` instead of a local re-declaration.

### 4. Test fat removed

- The W16 evidence test keeps replay, budget, and HTTP stickiness coverage but no longer re-runs the crash/resume compose arms that `composition_test.py` pins assertion-for-assertion (measured -61s), and its `_Judge` fixture stops scanning the whole store per judgment.
- Vacuous Protocol-machinery tests, fixture echoes, duplicate CLI surface pins, and duplicated journaled-behavior twins deleted; four byte-similar tamper tests collapsed to one parametrized case; verbatim micro fixture-builders consolidated onto canonical copies.

### Behavior notes (all verified untested-and-unpinned before changing)

- `--root` help text unified to "Local .wmo project root."
- A few error messages unified to the canonical helper's wording; no test or handler matched the old strings.
- Scalar judge responses with an empty dimensions array now fail via `ManualJudgeError` instead of a pydantic `ValidationError` (both `ValueError` subclasses).

### Gate

`uv run ruff check .` / `ruff format --check .` / `ty check` all pass; `uv run pytest -q` 1,240 passed, 1 skipped in 4:37 on the current tree, including the suites main added in #451-#457. The branch history preserves the remote rebase and its fixup as ancestors (no force pushes).


kfallah:
@greptileai review

kfallah:
bugbot run

kfallah:
@greptileai review

PR #503, opened 2026-08-19
Add deadline-aware async provider core

kfallah:
## What changed

- add an async-native provider transport backed by `httpx.AsyncClient`
- collapse transport and empty-output retries into one absolute-deadline attempt loop
- preserve one idempotency identity across safe same-endpoint retries
- keep existing synchronous `ModelClient` callers through a bounded compatibility adapter
- add typed, content-free provider failures and refusal signals for OpenAI, Anthropic, OpenAI-compatible, Gemini, and Bedrock
- add capability preflight, bounded blocking-worker admission for Bedrock, and explicit Tinker gateway exclusion
- make the runtime catalog construct the async transport by default while retaining deterministic sync transport injection

## Why

The provider stack was synchronous and applied separate full-budget retry loops at the HTTP and decoded-output layers. That prevented reliable cancellation and could multiply attempts without one request-wide bound. This change establishes the async execution and failure contracts needed by provider streaming while preserving current optimizer behavior.

## Scope boundaries

This PR does not implement public OpenAI SSE, cross-provider waterfalls, gateway CLI or lifecycle, SQLite state, or launch-provider streaming certification. Bedrock remains a bounded blocking compatibility lane, and Tinker remains optimizer-only.

## Validation

- `uv run ruff check .`
- `uv run ty check`
- 125 focused provider, gateway-contract, journal, model, and registry tests

Full repository pytest and installed-wheel certification continue in the review pass.

### DIFF
--- a/wmo/runtime/models/providers/__init__.py
+++ b/wmo/runtime/models/providers/__init__.py
@@ -1,8 +1,14 @@
 """Native provider clients and shared non-streaming transport behavior."""
 
 from wmo.runtime.models.providers.anthropic import AnthropicClient
+from wmo.runtime.models.providers.async_transport import (
+    AsyncJsonHttpTransport,
+    HttpxAsyncJsonTransport,
+    ProviderDeadlineExceeded,
+    RequestDeadline,
+)
 from wmo.runtime.models.providers.azure import AzureClient
-from wmo.runtime.models.providers.bedrock import BedrockClient
+from wmo.runtime.models.providers.bedrock import BedrockClient, BoundedBedrockClient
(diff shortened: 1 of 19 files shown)


kfallah:
Addressed the bounded Bedrock integration finding in 8e83486f: RuntimeModelCatalog now returns BoundedBedrockClient for completion while retaining the underlying BedrockClient only for embeddings. The bounded wrapper preserves synchronous optimizer callers and exposes deadline-aware complete_async(). Focused coverage exercises both paths; 29 provider/registry tests pass, and full Ruff/format/ty checks are green.

PR #474, opened 2026-08-17
Add the end-to-end build wizard

kfallah:
## Summary

- make bare `wmo build PROJECT` launch a defaults-first, resumable wizard that starts from an explicit upfront workflow picker (providers, build, judge rubric off by default, judge calibration off by default, router optimization), always prompts for an explicit local trace path when `-t/--traces` is absent (never discovers fixture files), configures providers and models, shows one exact bounded spend plan, and completes simulation, separate serving and fit RAG indexes, syllabus, closed-loop candidate evaluation, and router fit
- route explicit `-t/--traces` automation and the wizard through the same typed grounded-build seam while retaining the active hidden trace positional used by packaged consumers
- reuse the existing automatic router composition service; historical traces contribute router cells only with trustworthy candidate identity, while every missing task-candidate cell gets a fresh configured candidate rollout and judge score
- keep provisional judgment as truthful immutable provenance without blocking activation: the default router loads through `wmo run`, and optional human calibration followed by build creates a distinct human-calibrated successor
- add a pure automatic-router cost plan and enforce exact pre-consent and post-build equality before paid router work

## Compact TUI rework (latest commits)

- open with a cyan-bordered "Let's create a world model" welcome panel and terse one-line steps throughout the wizard and manual build path
- providers-first setup: pick providers, then assign each model role in turn (world model, judge, embedder, optional router candidates plus incumbent), with eligible models sorted so the maintained recommendation for the role comes first

```text
World model  (up/down + Enter; / search, b back, q cancel)
 ❯ gpt-5-6-luna
   gpt-5-6-terra
   ... 3 more below
```

- compact picker: alias-only role rows (no `(provider)` annotations), a bold cyan `❯` focus pointer, hints inline in the title, a visible window with dim `... more` indicators instead of showing every option at once, and providers capped at 3 visible rows; the picker renders inline below the command via a non-alternate-screen Rich `Live` region
- lowercase provider labels (`openai`, `anthropic`, `gemini`, ...) with no credential env-var status lines, and a one-line-per-item configuration summary with green checkmarks
- the pre-consent plan block, the `inspect` progress line, and the existing-catalog `providers` line are removed entirely: the only pre-spend output on the over-budget path is the concise `warning estimated $NN.NN exceeds the $MM.MM budget.` line plus `Proceed anyway? [y/n] (n)` (prices always two decimals, rounded up); the assumptions line, cost preflight, in-budget warning, resume text, calibration note, and router-plan explanation are all gone
[...]

PR #522, opened 2026-08-19
Delete dead lanes and fix confirmed happy-path bugs (net -11,510 non-test lines)

kfallah:
Full-repo review pass focused on the two happy paths, followed by a net-negative refactor. Every deletion was adversarially verified (zero production callers, checked against string imports, lazy export tables, docs, release pins, workflows) before it landed. Every fix reproduces a confirmed defect on the happy paths.

**Line accounting vs main: +562 / -12,071 = net -11,509 non-test Python lines.** Gates: ruff check, ruff format, ty, the full suite (1760 passed, 2 skipped), and the installed-wheel release evidence driver are green.

## Deletions (six clusters, one commit each)

- **Hosted noninteractive router lane (~6.4k lines).** `run_hosted_router_workflow`, `hosted_preflight/verification/spend/stages`, `attempt_authority`, `provisional`, `spend.py`, `common/project/{bundle,hosted_state,events,catalog}`, plus follow-on trims in live files. Zero callers in this repo or sibling repos. It duplicated the live `compose_router`/CLI orchestration for a consumer that does not exist. `docs/reference/router_optimization_config.md` rewritten around the live path; `project-bundles.md` deleted.
- **Legacy journaled-completion stack (~2.2k lines).** `RouterRuntime.complete()/select()`, `completion.py`, `journal_service.py`, the journal write side, and `runtime_selection.py`. Nothing writes the interaction journal in production; the gateway is selection-only by design. The gateway's selection bridge is now the public API (`reuse_sticky_selection`, `select_unretained`, `retain_prepared_selection`), removing the private-member pokes. The journal read surface for the SFT lane is untouched.
- **Runtime-RAG-refresh lane and PostHog network pull (~1.7k lines).** The refresh lane consumed a journal nothing writes. The CLI PostHog file source keeps working.
- **Gateway small surfaces (~620 lines).** Certification matrix module, dead `env://` secrets seam, unused re-export table, superseded catalog-authority functions, write-only counters and fields, unused pepper rotation.
- **Provider/protocol small surfaces (~560 lines).** Sync POST transport surface, migration shims, test fakes moved out of production modules, dead protocol adapters.
- **Common/CLI small surfaces plus manifest flattening (~420 lines).** Field manifests collapsed to accepted-field sets with byte-identical public errors; dead authoring module, wrappers, and vacuous helpers.
- **27 vacuous test stubs deleted** (zero bytes or a lone docstring). Note: AGENTS.md rule 2 currently blesses empty sibling test files; this PR removes the empty ones per your instruction, so rule 2's wording may deserve an update.

## Fixes (four clusters, all with regression tests that fail pre-fix)

Gateway serving:
- Streamed tool calls on Gemini and Bedrock crashed the public SSE encoders (missing `tool_call_index`); empty-argument tool calls crashed on Bedrock and Anthropic.
- Refusal failover dropped the provider's refusal text when no fallback was claimable; callers got a bare failure.
- A keyed stream that ended FAILED was cached and replayed for 24h, so retries with the same `Idempotency-Key` could never succeed.
- Joiners of an in-flight keyed request waited with no bound; now bounded by the request budget.
- `X-Client-Request-Id` no longer acts as an idempotency key; replay is `Idempotency-Key` only.
- Responses `instructions` were folded into stored history, so continuations inherited and accumulated them, and the envelope always echoed `instructions: null`. Instructions are now first-class: applied at head-of-context, echoed in the envelope, never inherited via `previous_response_id`.

Gateway throughput (the single point of failure):
- All SQLite authorization/accounting writes ran synchronously on the event loop; they now run off-loop. The pepper file was re-read and re-parsed per candidate row per request; now cached by inode/mtime/size with the permission check still on every call.
- Every provider call opened a fresh TCP+TLS connection; the gateway now owns one pooled `httpx.AsyncClient`, closed after drain.
- Keyed-request replay claims were an O(4096) scan under one lock; now an O(1) index. Streaming paid a full pydantic JSON serialization per token for byte accounting and retained Chat events that nothing consumed; both removed.

Usage/budgets and build path:
- `gateway status` compared key expiry lexically against a local-offset timestamp; wrong for hours around expiry on any non-UTC host.
- Crash reconciliation never released monthly-budget reservations; budgets leaked permanently after a crash.
- A typo'd `--pool`/`--deployment` budget scope was accepted and matched nothing (fail-open); now validated against the active revision.
- Two operator errors named a CLI flag that does not exist; OTLP ingest dropped span failures so failed traces carried no outcome; the prospective router-candidate catalog dropped `candidate_reasoning_efforts`.
- `wmo config gateway --help` hid the `pool` and `budget` groups; `wmo run` crashed on closed stdin instead of emitting the stable receipt.

## Kept deliberately

- `load_router` (embeddable entry): release-pinned public replay surface.
- Sandbox engine and `simulation/comparison.py`: release-evidence pins (W16) in release tests and CI workflows, and AGENTS.md names them owned surfaces.
- `ModelRoles.rubric_proposer/teacher`: removal would fail-closed every persisted `models.toml`.

## Needs your call (found in review, not changed here)

1. Root README still shows `router.complete()` in "Using the API"; that surface is gone. README edits need your explicit approval, say the word and I will fix it.
2. The closed parameter profile 400-rejects official OpenAI params (`top_p`, `user`, `seed`, penalties, `logprobs`, `n`, `store`). Stock clients that send them fail hard. Supporting at least `top_p` everywhere is a small mapped change; the rest are OpenAI-family only.
3. Responses continuations and idempotency replay are process-local; a restart 400s valid `previous_response_id` chains and 409s keyed retries. Durability here is a product decision.
4. `wmo optimize model` (SFT) reads a runtime journal that nothing writes anymore; the lane only works on legacy journals. Either the gateway learns to journal routed project traffic or the lane needs a new evidence source.
5. Budget reservations over-estimate roughly 5-10x (bytes as tokens, rates summed), so monthly limits reject far below the real cap.
6. Telemetry defaults on with a hardcoded PostHog key; worth an explicit launch decision for a local BYOK gateway.
7. Smaller items are listed in the review notes: `--check` conflicts with a running gateway's instance lock, ready receipt prints before bind, `supports_streaming=False` deployments can serve nothing, CLI accepts trace counts the Python API rejects, sticky-episode eviction under multi-tenant load, duplicate embedding dispatch without single-flight.

## Verification

- Full E2E on this branch with a loopback mock provider: per-key model listing, 401/403 isolation, Chat and Responses with and without streaming, continuation, idempotent replay, per-identity usage attribution, and `wmo build` over a 280-trace OTLP corpus (70 tasks, serving+fit RAG, grounded world model, 50/20 split).
- Same drives were run on main before the change as the baseline; the instructions-echo and non-inheritance behaviors are visible improvements.
- Branch is rebased on current main three times as main moved: over the CLI subpackage restructure and tests-directory moves (#517/#519), over the end-to-end build wizard (#474), and over the four new gateway seams (#523 embeddable runtime factory, #525 pluggable protocol state, #527 immutable project activation, #529 platform-neutral contracts). All four seams are fully preserved and my fixes were re-fitted onto them: the pooled HTTP client now closes through the composition runtime's flush hook, the joiner timeout applies to the pluggable replay lease, and the O(1) replay index rides the new store internals. Two reconciliation decisions worth knowing: the decision-sink recording I had deleted as dead is now live in #527's activation seam, so it stays; and #527's new activation test was ported from the removed RouterRuntime.select() to the public selection trio.


kfallah:
@greptile-apps review

kfallah:
@greptile-apps review

kfallah:
@greptile-apps review

kfallah:
@greptile-apps review

PR #533, opened 2026-08-20
Keep Unix-only stdlib out of CLI and pytest import paths

### LINKED ISSUE #532: pytest cannot collect on Windows: wmo/conftest.py imports Unix-only fcntl
## Summary

On Windows, `uv run pytest` fails during collection, before any selected tests run. Pytest loads `wmo/conftest.py`, which imports `fcntl` at module scope. `fcntl` is a Unix-only CPython stdlib module, so collection dies with `ModuleNotFoundError`.

This is not a missing PyPI dependency. `uv sync --extra dev` succeeds. The crash is an unconditional POSIX import in a suite-wide conftest.

There is a second, independent import barrier: even with `--noconftest`, `wmo/optimize/router/composition_test.py` still fails because `wmo.cli.shared.picker` imports `termios` at module scope, and CLI startup pulls that in.

Windows is not covered by CI (`ubuntu-latest` plus one macOS evidence job). `pyproject.toml` also has no `Operating System ::` classifiers, so the published wheel still looks platform-agnostic. Related history: #345 made production locking portable via `filelock`; #359 / #360 asked whether Windows was a target and were closed without a stated policy.

This is not a router-composition regression. Router library modules still import. Collection never reaches those tests.

## Environment

- OS: Windows 11 (`win32`)
- Python: 3.13.2 (64-bit)
- Install: `uv sync --extra dev` (editable `world-model-optimizer 0.3.0`, pytest 9.1.1)

Confirmed absent on this interpreter:

- `fcntl`: `find_spec` is `None`
- `termios`: `find_spec` is `None`
- `pty` / `tty`: source files exist, but importing them fails because they depend on `termios`

Confirmed still importable:

- `wmo.common.core.locks`
- `wmo.optimize.router.composition`

Confirmed not importable:

- `wmo.cli.app` (`picker.py` -> `termios`)

## Reproduction

```bash
uv sync --extra dev
uv run python -c "import fcntl"
# ModuleNotFoundError: No module named 'fcntl'

uv run pytest -q wmo/optimize/router/composition_test.py wmo/optimize/router/composition_evidence_test.py
```

Actual:

```text
ImportError while loading conftest '...\wmo\conftest.py'.
wmo\conftest.py:6: in <module>
    import fcntl
E   ModuleNotFoundError: No module named 'fcntl'
```

Exit code 4 (collection error). No tests ran.

Skipping the conftest still fails:

```bash
uv run pytest --noconftest -q wmo/optimize/router/composition_test.py
```

```text
wmo/optimize/router/composition_test.py
  -> wmo.cli.build.app
    -> wmo.cli.providers.provider_picker
      -> wmo.cli.shared.picker
        import termios
E       ModuleNotFoundError: No module named 'termios'
```

## Expected

One of these should be true, and the repo should say which:

1. **POSIX-only:** document that pytest and the CLI require POSIX, add OS classifiers, and treat this traceback as expected.
2. **Windows should at least collect unit tests / import the CLI:** Unix-only APIs should not be imported until a POSIX-only fixture or raw-terminal path actually needs them.

Today neither is stated, so a Windows contributor hits a hard collection failure on tests that do not use PTYs.

## Why collection fails

`wmo/conftest.py` is a directory-level conftest. Pytest imports it before collecting anything under `wmo/`, including router unit tests. The file exists for interactive CLI tests (`terminal_child` / `python_terminal_child`): it opens a PTY, sets window size with `fcntl.ioctl`, and drives keys. Those helpers are Unix-only, but the imports run at module load rather than inside the fixture.

Hatch already excludes `wmo/conftest.py` from the wheel (`exclude = ["**/*_test.py", "wmo/conftest.py"]`). Local pytest still loads it from the source tree.

Other POSIX-only module-scope imports that will fail next if collection gets past conftest:

- `wmo/cli/shared/picker.py` (`termios`, `tty`)
- `wmo/cli/judge/trace_viewer.py` (`termios`, `tty`)
- `wmo/runtime/environments/local_test.py` (`fcntl`)
- `wmo/tests/release_test.py` (`pty`, `termios`)

## Proposed direction

This should not try to vendor `fcntl`, add a Windows PTY layer, or make local-process / agent-shell execution work on Windows.

Suggested split:

1. **Decide and record the platform policy** (POSIX-only vs library/CLI importable on Windows).
2. **If pytest should collect on Windows:** keep the PTY helpers, but import `fcntl` / `pty` / `termios` inside `_run_terminal_child`, and `skipif(os.name != "posix")` the terminal fixtures. Apply the same skip/lazy-import pattern to `local_test.py` and PTY paths in `release_test.py`.
3. **If `import wmo.cli` should work on Windows:** move `termios` / `tty` imports in `picker.py` and `trace_viewer.py` into the raw-key helpers. The picker already falls back to line-based input when stdin is not a TTY; Windows would use that path. Do not rewrite keyboard mode on `msvcrt`.

Interactive PTY tests can stay Unix-only. They just should not block unrelated tests from collecting.

## Out of scope

- Making sandbox / local-process / LLM-authored Unix shell execution work on Windows
- Changing router fitting, evidence, or composition behavior
- Adding a red `windows-latest` CI job before the import boundaries above are fixed

brainRottedCoder:
## Summary
- Stop importing `fcntl` / `pty` / `termios` / `tty` at module scope so pytest can collect on Windows and `import wmo.cli.app` works.
- Keep PTY child runners and Darwin FD probes POSIX-only via skip/local import.
- Add an AST and subprocess import gate so the boundary cannot regress.
Fixes #532
## Test plan
- [ ] `uv run ruff check .`
- [ ] `uv run ruff format --check .`
- [ ] `uv run ty check` (Ubuntu CI; Windows ty still fails on POSIX stubs)
- [ ] `uv run pytest -q wmo/tests/posix_stdlib_import_test.py wmo/cli/shared/picker_test.py wmo/cli/judge/trace_viewer_test.py`
- [ ] `uv run python -c "import wmo.cli.app; import wmo.conftest"`
- [ ] `uv run pytest --collect-only -q wmo`

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -148,6 +148,10 @@ uv run pytest -q
   (`logging.getLogger(__name__)`), never the `print` builtin — enforced by ruff's `T20` rules.
   The one exception is deliberate user-facing CLI presentation, which goes through a local rich
   `Console` owned by the command module (that is product output, not logging).
+- Library import, CLI import, and pytest collection must not import Unix-only stdlib modules
+  (`fcntl`, `pty`, `termios`, `tty`) at module scope. Load those modules only inside the POSIX
+  helpers that need them. Pseudo-terminal child runners and Darwin file-descriptor probes stay
+  POSIX and skip elsewhere.
 
 ## Writing
 
--- a/exp/cli/judge/trace_viewer.py
+++ b/exp/cli/judge/trace_viewer.py
@@ -12,8 +12,6 @@ from __future__ import annotations
 
 import json
 import sys
-import termios
-import tty
 from collections.abc import Callable, Iterator
 from contextlib import contextmanager
 from dataclasses import dataclass
@@ -558,6 +556,9 @@ def _read_key() -> str:
     Returns:
         Single character, or a normalized name such as ``up`` or ``pgdn``.
     """
+    import termios
+    import tty
+
     descriptor = sys.stdin.fileno()
     saved = termios.tcgetattr(descriptor)
     try:
--- a/exp/cli/shared/picker.py
+++ b/exp/cli/shared/picker.py
@@ -14,8 +14,6 @@ from __future__ import annotations
 import os
 import select
 import sys
-import termios
-import tty
 from collections.abc import Callable, Iterator, Sequence
 from contextlib import contextmanager
 from dataclasses import dataclass
@@ -293,9 +291,11 @@ def uses_keyboard_list(console: Console) -> bool:
         console: Terminal used for the screen.
 
     Returns:
-        True only when both stdout and stdin belong to an interactive terminal.
+        True only on POSIX when both stdout and stdin belong to an interactive terminal.
+        Windows consoles can report a TTY without Unix terminal I/O, so they use the
+        line-based path.
     """
-    return console.is_terminal and _stdin_is_tty()
+    return os.name == "posix" and console.is_terminal and _stdin_is_tty()
 
 
 def interpret_key_bytes(data: bytes) -> PickerEvent:
@@ -436,6 +436,9 @@ def _terminal_key_reader(
     if read_key is not None:
         yield read_key
         return
+    import termios
+    import tty
+
     fd = sys.stdin.fileno()
     old = termios.tcgetattr(fd)
     try:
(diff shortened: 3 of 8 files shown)


kfallah:
thanks for the contribution but this feature isn't currently planned

brainRottedCoder:
okk!

PR #538, opened 2026-08-20
gateway: tolerate reasoning content in Anthropic and Gemini native streams

SilenNaihin:
**Draft for Kion's review as module owner. Do not merge.** This is a targeted fix for a first-party direct-serving bug the platform team found on the live gateway. Non-blocking (first-party currently serves via its OpenRouter fallback rung), but this is the fix that lets the platform route first-party onto direct provider keys.

## The bug

The gateway streams natively to first-party providers and normalizes their SSE. Both native stream parsers reject any content shape they do not recognize by raising `ProviderResponseError`, which `normalized_provider_failure` classifies as `malformed_response` (`failover_eligible`). Current frontier models emit reasoning content that is not gateway-visible output, so a clean provider `200` was surfaced as `malformed_response` and the model fell through to its fallback rung.

- **Anthropic** (`_anthropic_messages_events` in `streaming.py`): a `content_block_start` whose block type was not `text`/`tool_use`/`refusal` raised "unsupported content block", and a `content_block_delta` whose delta type was not `text_delta`/`input_json_delta`/`refusal_delta` raised "unsupported content delta". Extended-thinking responses send a `thinking` block with `thinking_delta` and `signature_delta` deltas, which tripped both.
- **Gemini** (`_gemini_events` in `gemini_streaming.py`): a part carrying neither non-empty `text` nor a `functionCall` raised "unsupported part", and an empty-string text part raised "text part must be text". Gemini 2.5 emits `thought` parts and bare `thoughtSignature` parts, which tripped this.

## The fix

Both parsers now skip reasoning content instead of rejecting it, matching how robust clients accumulate only visible output:

- Anthropic: unrecognized `content_block_start` / `content_block_delta` are skipped; `text`, `tool_use`, and `refusal` are unchanged. A skipped thinking block registers no tool accumulator, so its `content_block_stop` is already a no-op.
- Gemini: a part flagged `thought`, and any part with neither visible text nor a `functionCall` (for example a bare `thoughtSignature`), is skipped; empty text parts are skipped. A non-string `text` value is still a malformed response.

## Tests

Colocated regression tests (whole-repo gate green: `ruff check`, `ruff format`, `ty check`, `pytest` over `exp/runtime/models/providers/` and `exp/runtime/gateway/`, 292 passed):

- `test_anthropic_stream_skips_extended_thinking_blocks_and_deltas`: a stream with a `thinking` block plus `thinking_delta` and `signature_delta`, then visible text, yields exactly `TEXT_DELTA`, `USAGE`, `COMPLETED`.
- `test_gemini_stream_skips_reasoning_parts_and_thought_signatures`: a stream interleaving a `thought` part, a bare `thoughtSignature`, and an empty text part with visible text yields exactly `TEXT_DELTA`, `USAGE`, `COMPLETED`.

## Note on the third symptom (OpenAI `provider_authentication`)

The platform also saw `provider=openai` direct return `provider_authentication`. That is a genuine upstream 401/403 (`stream_attempts.py` raises `ProviderTransportError(status_code)` which `_transport_failure` maps to `PROVIDER_AUTHENTICATION`). The OpenAI adapter authenticates correctly (`Authorization: Bearer <key>`, key read from the connection's `api_key_env`, posting to `/v1/responses`). No adapter change is warranted; this is being flagged to the platform side as a house-connection credential/wiring question (the value bound to the OpenAI connection's `api_key_env`, or that key's Responses-API access), not a WMO bug.

### DIFF
--- a/exp/runtime/models/providers/gemini_streaming.py
+++ b/exp/runtime/models/providers/gemini_streaming.py
@@ -114,9 +114,9 @@ async def _gemini_events(sse: _SseDecoder) -> AsyncIterator[GatewayEvent]:
             )
             for raw_part in parts:
                 part = require_object(raw_part, "Gemini candidate part")
-                text = part.get("text")
-                if isinstance(text, str) and text:
-                    yield factory.create(GatewayEventKind.TEXT_DELTA, text_delta=text)
+                if part.get("thought") is True:
+                    # Reasoning parts (thought text and thought signatures) are not
+                    # gateway-visible output, so they carry nothing to surface.
                     continue
                 if part.get("functionCall") is not None:
                     async for event in _tool_events(
@@ -127,9 +127,16 @@ async def _gemini_events(sse: _SseDecoder) -> AsyncIterator[GatewayEvent]:
                         yield event
                     tool_index += 1
                     continue
+                text = part.get("text")
+                if isinstance(text, str):
+                    if text:
+                        yield factory.create(GatewayEventKind.TEXT_DELTA, text_delta=text)
+                    continue
                 if text is not None:
                     raise ProviderResponseError("Gemini text part must be text")
-                raise ProviderResponseError("Gemini stream emitted an unsupported part")
+                # A part with neither visible text nor a function call (for example a
+                # bare thought signature) carries no gateway-visible output; skip it.
+                continue
         finish_reason = candidate.get("finishReason")
         if finish_reason is None:
             continue
--- a/exp/runtime/models/providers/gemini_streaming_test.py
+++ b/exp/runtime/models/providers/gemini_streaming_test.py
@@ -216,6 +216,55 @@ def test_gemini_stream_cancellation_closes_the_active_http_response() -> None:
     asyncio.run(scenario())
 
 
+def test_gemini_stream_skips_reasoning_parts_and_thought_signatures() -> None:
+    """Thought parts and bare thought signatures are skipped, not rejected as malformed."""
+
+    async def scenario() -> None:
+        """Consume a stream that interleaves reasoning parts with visible text."""
+        upstream = _ChunkStream(
+            (
+                _sse(
+                    {
+                        "candidates": [
+                            {"content": {"parts": [{"text": "pondering", "thought": True}]}}
+                        ]
+                    }
+                ),
+                _sse({"candidates": [{"content": {"parts": [{"thoughtSignature": "sig-abc"}]}}]}),
+                _sse({"candidates": [{"content": {"parts": [{"text": ""}]}}]}),
+                _sse({"candidates": [{"content": {"parts": [{"text": "answer"}]}}]}),
+                _sse(
+                    {
+                        "candidates": [{"finishReason": "STOP"}],
+                        "usageMetadata": {"promptTokenCount": 4, "candidatesTokenCount": 2},
+                    }
+                ),
+            )
+        )
+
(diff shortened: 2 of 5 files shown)


SilenNaihin:
### Added commit: guard the legacy `max_tokens` -> native Responses translation (no behavior change)

The platform team relayed a second reported first-party-direct symptom: a client sending `max_tokens` to a GPT model on the direct `/v1/responses` path getting a 400 "Unknown parameter: 'max_tokens'" (surfaced as `provider_internal`).

I traced the gateway request path and the WMO serving layer already handles this correctly, at both the pinned rev (`cb4c1607`) and `main`:

- `decode_chat` maps `max_completion_tokens or max_tokens` -> canonical `maximum_output_tokens` (`openai_protocol/requests.py`).
- `openai_responses_stream_payload` emits `max_output_tokens` only, and never `max_tokens` (`models/providers/streaming_requests.py`).
- `_WireModel` is `extra="forbid"` and there is no raw field passthrough anywhere, so a stray `max_tokens` sent to the `/v1/responses` surface is rejected at the gateway boundary and never reaches OpenAI.

So no behavior fix is warranted in WMO. The added commit is a regression **guard** only: existing coverage exercised `max_completion_tokens` but not the legacy `max_tokens` field, so this decodes a chat request carrying `max_tokens` and asserts the native Responses payload carries `max_output_tokens` and never `max_tokens`.

Because WMO cannot emit `max_tokens` to `/responses`, the observed OpenAI 400 must originate on a path outside this gateway (for example a verification harness calling OpenAI directly, or a platform-side direct call). Flagging that for the platform side rather than patching around it here.

SilenNaihin:
### Correction + definitive evidence on the max_tokens / gpt-5.5 report

Resolved with hard data (supersedes my earlier "originates outside the gateway" note).

**Running worker commit:** the live everything-preview gateway-worker has WMO `cb4c1607` installed (`world_model_optimizer-0.3.0.dist-info/direct_url.json` -> `commit_id: cb4c1607...`). Not a stale image.

**Raw OpenAI `/v1/responses` probes (house key, gpt-5.5):**
- `max_output_tokens: 8` -> 400 `invalid_request_error` param `max_output_tokens`: "integer below minimum value. Expected a value >= 16, but got 8."
- `max_output_tokens: 2000` -> 200 completed.
- `max_tokens: 8` -> 400 `invalid_request_error` param `max_tokens`: "Unknown parameter: 'max_tokens'."

**Conclusion:** the gateway translation at cb4c1607 is correct. A caller's `max_tokens` and `max_completion_tokens` both map to canonical `maximum_output_tokens` and reach OpenAI as `max_output_tokens` (200 at 2000). The through-gateway 502/`provider_internal` repro used `max_tokens: 8`; that becomes `max_output_tokens: 8`, which is below gpt-5.5's reasoning-model minimum of 16, so OpenAI 400s. `max_completion_tokens: 8` fails identically for the same reason (both correctly become 8), and omitting the limit uses a default above 16 so it 200s. The earlier "Unknown parameter: 'max_tokens'" was a raw call sending the wrong field straight to `/responses` (probe 3), not the gateway path.

So there is no field-name bug in WMO, and no fix is warranted for it. The guard test in this PR still stands (it pins the correct legacy-max_tokens -> max_output_tokens contract). The 8-token failure is a caller/test-value artifact (a value below the model's minimum), not a WMO or platform serving bug.

kfallah:
@greptile review

PR #540, opened 2026-08-20
gateway: surface invoked tool names on per-attempt usage

SilenNaihin:
**Draft for Kion's review as module owner. Do not merge.** Third first-party/gateway fast-follow in the same pin-bump change set as #538 (anthropic/gemini reasoning-parse) and #534 (accounting latch). Non-launch-blocking: the platform half is already built and merged and stays defensive-NULL / honest empty state until this lands and the pin bumps.

## Goal

Let the gateway surface the tool NAMES a model actually invoked on the per-attempt usage the platform settles, so the "Tools called" telemetry lights up. Names only, never arguments (the platform ledger is content-free by contract).

## Change

- `contracts.py`: `GatewayUsage` gains `tool_names: tuple[str, ...] = ()` (first-use order, names only, documented). Default `()` keeps every existing construction and the platform's non-bumped path byte-identical.
- `execution.py`: `GatewayExecutionStream` accumulates invoked names per physical attempt. As each `TOOL_CALL_COMPLETED` passes, its `tool_call.name` is appended to that attempt's list, deduplicated in first-use order (right where `latest_usage` is already tracked). A single `_stamp_tool_names` helper stamps the accumulated names onto the terminal event's usage on both the settled-success path and the `abort` path, mirroring how token counts flow from attempt to event. Semantics are INVOKED tools (what the model called), not `request.tools` (offered).

Per-attempt is correct: the platform copies the winning (last) attempt's `tool_names` onto the event, the same way `input`/`output_tokens` flow attempt -> event.

## Platform activation (not here; happens at the pin bump)

The one-liner is already staged on the platform side (`explabs/gateway/ledger.py` `finish_attempt`, gw/int-p4-worker / #472): the marked `None,` becomes `None if usage is None else list(usage.tool_names)`. `list(...)` on the `tuple` field works directly. Storage/read/display already merged (schema #456, reads #463, FE #490).

## Tests

Colocated in `exp/runtime/gateway/tests/data_plane_test.py`:
- a stream invoking two tools (one repeated by name) settles terminal usage with `("search", "lookup")` (deduplicated, first-use order), driven through the full service with a separate USAGE event then COMPLETED (the real provider ordering).
- a tool-free stream settles with `()`.

Whole-repo gate green: `ruff check` + `ruff format`, `ty check`, `pytest` over gateway + providers + openai_protocol (349 passed).

### DIFF
--- a/exp/runtime/gateway/contracts.py
+++ b/exp/runtime/gateway/contracts.py
@@ -171,14 +171,36 @@ class GatewayRequest(ContractModel):
 
 
 class GatewayUsage(ContractModel):
-    """Normalized token counts and invoked tool names from one provider attempt."""
+    """Normalized token counts and invoked tool names from one provider attempt.
 
-    input_tokens: int = Field(ge=0)
-    output_tokens: int = Field(ge=0)
+    A terminal event may carry only ``tool_names`` when the provider omits token usage. In that
+    case both token totals remain unknown instead of being represented as zero.
+    """
+
+    input_tokens: int | None = Field(default=None, ge=0)
+    output_tokens: int | None = Field(default=None, ge=0)
     cached_input_tokens: int | None = Field(default=None, ge=0)
     reasoning_tokens: int | None = Field(default=None, ge=0)
     tool_names: tuple[str, ...] = ()
     """Invoked tool names in first-use order, names only and never arguments."""
+
+    @model_validator(mode="after")
+    def _require_complete_tokens_or_tool_names(self) -> GatewayUsage:
+        """Require complete token totals unless this is tool-only terminal metadata."""
+        totals = (self.input_tokens, self.output_tokens)
+        if (totals[0] is None) != (totals[1] is None):
+            raise ValueError("input and output token counts must be reported together")
+        if totals[0] is None:
+            if self.cached_input_tokens is not None or self.reasoning_tokens is not None:
+                raise ValueError("token detail counts require input and output totals")
+            if not self.tool_names:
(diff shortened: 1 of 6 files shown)


SilenNaihin:
### Exact platform activation ref (confirmed with the telemetry-tools owner)

The platform reads whatever `GatewayUsage` the terminal event carries; it does not itself look back at prior USAGE events. That is why this PR stamps `tool_names` onto the terminal's carried usage (whether inherited from a prior USAGE event via `_with_latest_usage` or set directly on the terminal): it survives to the platform in both provider orderings.

Activation at the pin bump (no platform change until then; stays defensive-NULL):
- `explabs/gateway/ledger.py`, `PostgresAttemptLedger.finish_attempt` (branch `gw/telemetry-tools-worker`, commit `3cb3237f`, folds onto int-p4 / #472).
- The 10th arg of `select public.gateway_settle_attempt(%s x10)` is a marked defensive `None,`; it flips to `None if usage is None else list(usage.tool_names)`. `usage` is the `GatewayUsage` from `_terminal_values`. `list()` on this PR's `tuple[str, ...]` field works directly.
- `gateway_settle_attempt` collapses an empty `p_tool_names` to NULL, so the default `()` reads as the same honest "called no tools" empty state.

Write path proven green by pgTAP on the schema car (#456, `supabase/tests/database/explabs_gateway_usage_tools.test.sql`, 10/10): settle persists `p_tool_names` -> `attempt.tool_names` -> finalize copies to `event.tools_used`.

kfallah:
@greptile review

kfallah:
@greptile review

PR #543, opened 2026-08-20
Use builder provider selector in first-run gateway setup

kfallah:
## Summary

- Reuse the builder's shared provider selector for first-run gateway setup while persisting every selected provider connection.
- Show the provider model, exact logical model ID, public alias, and identity ID together with sensible defaults.
- Accept all defaults with one Enter; type edit to change the fields.
- Replace mutation-plan and confirmation noise with one progress display and an Experiential-green completion state.
- Keep non-interactive provider overrides and the singleton first-run gateway bootstrap intact.
- Keep `exp run` startup output limited to the green ready state and the exports needed to use the gateway.
- Abort cleanly on selector or shared metadata prompt cancellation and retry invalid custom credential environment names.

## Validation

- Focused build/config/gateway/run pytest suite: 135 passed.
- Ruff check and format check passed.
- ty check passed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,8 +16,8 @@ Experiential is an open source gateway and router for agent workflows:
 
 ## Getting Started
 
-Start a local OpenAI-compatible gateway. On first run, the setup wizard asks for a provider,
-model, and public alias, then prints a one-time virtual key:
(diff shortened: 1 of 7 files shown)


kfallah:
@greptileai review

PR #592, opened 2026-08-22
Fix main's red gate: dead-fallback probe uses a no-dialect alias

kfallah:
Main's gate is failing on `test_dead_fallback_engine_degrades_only_escalated_routes` at both the #587 and #576 merge points: the test probed `/v1/responses` expecting escalation to the fallback engine, but #582/#585 made Responses and project aliases natively served, so the fixture alias never escalates anymore.

The probe now targets a second seeded alias backed by a provider with no native dialect (gemini), which the admission path escalates by construction, so the test stays valid no matter which public surfaces become native next. Verified 3x consecutive green locally plus the full gateway suite.

This unblocks every PR that runs the gate against current main, including #588.


kfallah:
Superseded: the fix commit was cherry-picked into #588 (merged as a4cabb21) to unblock its gate in parallel, so this diff is now empty against main. Main's gate is green again.

PR #611, opened 2026-08-23
Add identity-scoped, default-off gateway guardrails

kfallah:
Identity-scoped, default-off gateway guardrails with an opt-in standard pack, plus interactive Setup Gateway authoring.

## Guardrail engine

- Policy lookup is `(organization_id, identity_id)`. The local organization is `local`.
- Default-off: missing `ROOT/gateway/guardrails.json` leaves the exact unguarded hot path (no engine). A valid file whose `policies` list is empty is also unguarded.
- Capabilities: `pii`, `secret_leakage`, `prompt_injection`, `content_safety`. Prompt injection is input-only.
- Actions: `allow`, `modify`, `block`, `error`.
- Input runs after Responses continuation and before route, preflight, accept, and waterfall. Output runs once on the winning completion. Tool-call arguments are blocked, not rewritten.
- Protected identities fail closed. Unprotected identities skip uncertain checks.
- Unique `check_id` only. Manual chains may repeat the same stage and capability.
- Async inspects run on bounded isolation workers (daemon threads with private event loops). The caller returns at the check timeout even if an adapter blocks before its first await. A worker stays occupied until that isolated invocation exits. An abandoned adapter is quarantined until every abandoned inspect for it finishes. `http_json` keep-alive reuse stays per isolation loop.
- Native callbacks submit enforcement onto one shared daemon loop so a Rust worker can return while an abandoned inspect still occupies an isolation worker.
- Leftover synchronous adapters run only through `BoundedSyncClassifier` on a fixed set of daemon workers. Admission is capped at `max_workers`. A full wrapper fails immediately instead of queueing work that timeout cannot cancel.
- Native crate is 0.1.5. The application version stays 0.5.4.

## Standard pack and http_json

- Never enabled globally. An identity opts in with `preset: standard`, an explicit `protected` boolean, and a binding for every capability.
- Presence-strict load: `preset` plus `checks` is always rejected. `capability_adapters` without `preset` is rejected. Null or empty `preset` is rejected.
- Documented expansion order with a 250 ms default per check, not 250 ms for the whole chain.
- `http_json` posts to a dedicated classifier URL. Optional `bearer_env` is an environment-variable name only.
- One capability-aware adapter may be bound to every required capability because the outbound contract includes `capability`.

## Interactive setup

- First-run Setup Gateway shows `Guardrails: Off`. Enter does not create `gateway/guardrails.json`.
- Edit can opt the selected identity into the standard pack: classifier URL plus optional bearer environment-variable name. Setup authors `protected=true` and the documented defaults for `manager.organization_id` plus that identity.
- One setup-owned `http_json` adapter is bound to all four capabilities. Setup-owned IDs are `stable_id` values over a structured `{organization_id, identity_id}` subject, with distinct `setup-http-json` and `setup-standard` prefixes. They do not concatenate raw fragments, so hyphenated pairs cannot collide and long valid identifiers stay bounded.
- Reconfiguration preserves the current file unless the operator changes it. A policy that is not the setup-owned standard pack is shown as `Custom/preserved` and is not replaced unless the operator types `replace`. Other identities, policies, adapters, and unrelated top-level fields stay unchanged.
- Config read-modify-write is locked and atomic. One non-reentrant lock covers snapshot, selected mutation, the rest of the setup mutation window, and proven-failure restoration. `AliasActivationOutcomeUnknownError` keeps the selected file. Standalone apply still takes the same lock.
- A symlink is written through and never removed. Disabling the last setup-owned pack writes an empty valid document through that link. A regular file created solely by setup is deleted when it becomes empty.
- Never persists prompt, response, or secret literals.

## Docs

- `docs/reference/gateway-guardrails.md` covers the interactive path, the empty-policies unguarded path, and the hand-authored file escape hatch.

## Out of scope

- No new root or config CLI commands.
- No management UI beyond this opt-in.
- World-model fidelity is unchanged and still has no authority over router fitting.

<div><a href="https://cursor.com/agents/bc-2b2d299c-ae69-4910-8d74-94064d72d337?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-2b2d299c-ae69-4910-8d74-94064d72d337&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


kfallah:
@greptileai

PR #615, opened 2026-08-23
feat(cli): sync Experiential models during login

kfallah:
## Summary

Follow-up to #610. `exp login` now:

- stores the endpoint-bound Experiential credential;
- lists every model returned for the authenticated account and registers `experiential-cloud` plus secret-free model records in the selected `.exp/models.toml`;
- preserves existing local aliases and active gateway deployment metadata;
- exposes Experiential Cloud in local gateway setup and marks selected hosted deployments `host_managed`.

`exp config providers` remains available for role assignment or explicit refresh.

## Verification

- `uv run ruff check .`
- `uv run ty check`
- `TERM=xterm-256color uv run pytest -n 4 --dist worksteal -q -k "not live" exp ...`: 2,285 passed, 1 skipped

The repository's terminal redraw tests require a non-`dumb` TERM; with the shell default `TERM=dumb`, those unrelated tests fail their terminal rendering assertions.

### DIFF
--- a/docs/reference/openai-compatible-recipes.md
+++ b/docs/reference/openai-compatible-recipes.md
@@ -126,9 +126,13 @@ authority for this path. Run `exp login` in an interactive terminal to open Plat
 store the returned key after the loopback callback completes.
 
 ```bash
-exp login
-exp config providers --provider experiential-cloud --root ROOT
+exp login --root ROOT
 ```
+
+Login registers the `experiential-cloud` connection and synchronizes every model identity returned
+by the hosted `/v1/models` endpoint for the authenticated account. Local gateway setup can select
+those models immediately. Use `exp config providers --provider experiential-cloud --root ROOT`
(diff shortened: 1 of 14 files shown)


kfallah:
@greptile review

Please review the current PR head `83eb028ebce1c2c01b6349950f4ff8991aee53a5` for the login-time Experiential Cloud provider and account-model synchronization change.

kfallah:
@greptile review

The 4/5 finding is fixed in `e246c17c`: login now rejects endpoint replacement when `experiential-cloud` has active gateway-owned records, and persists the credential only after synchronization succeeds. Please review the current PR head `e246c17c`.

kfallah:
@greptile review The second 4/5 finding is fixed in e4498ab3: login now passes active SQLite gateway provider authorities into synchronization, so an endpoint-changing login is rejected even when the catalog has no gateway metadata. The prior credential remains untouched. Please review the current PR head e4498ab3.

PR #628, opened 2026-08-25
fix(gateway): gate generation parameters by route capabilities

kfallah:
## Summary

- Enforce one exact generation-parameter contract across every deployment in a waterfall before any provider request opens.
- Cover `temperature`, `top_p`, `top_k`, `reasoning` and `reasoning_effort`, `max_tokens`, `max_completion_tokens`, `max_output_tokens`, logprobs, stop sequences, tools, tool choice, parallel tools, and structured output.
- Preserve the caller's public max-token field for field-specific errors, then translate accepted limits into each provider's required wire field.
- Validate model-specific numeric ranges, output ceilings, reasoning effort sets, configured effort pins, and route intersections locally. Unsupported or out-of-range semantic values return a public `400` and never reach a provider.
- Keep model defaults valid instead of silently clamping them. Examples include GPT-5 Pro defaulting to `high` and GPT-5.1 permitting `temperature` and `top_p` only when effective reasoning effort is exactly `none`.
- Translate accepted reasoning controls into native OpenAI Responses, OpenAI-compatible and Azure Chat, OpenRouter, Anthropic adaptive thinking, Gemini thinking-level shapes, and Vertex AI's publisher-scoped Gemini wire shape.
- Reject Anthropic thinking configuration/history and incompatible `tool_result.is_error` semantics instead of silently dropping them.
- Publish the resolved non-secret generation contract through `/v1/models`, and ingest provider-published capability metadata from OpenAI-compatible listings, OpenRouter, and Gemini.
- Require the provider certification matrix to cover every gateway-servable runtime provider. Tinker remains explicitly gateway-excluded.
- Bound provider-indexed reasoning-summary and tool-call stream state in Python and Rust. Empty reasoning fragments allocate no state, and hostile distinct indices fail at the shared retained-entry ceiling.

## Parameter policy

| Input | Behavior before dispatch |
|---|---|
| `temperature`, `top_p`, `top_k` | Forward only when every route rung supports the field and exact value range; otherwise local `400`. |
| `reasoning_effort` / `reasoning.effort` | Require the exact effort to exist in every rung's model-specific set; never omit or clamp an explicit value. |
| `max_tokens`, `max_completion_tokens`, `max_output_tokens` | Preserve the caller alias, enforce the smallest route ceiling, and translate to the provider-required field. |
| `logprobs=true` | Local `400` until normalized responses can preserve logprob results. |
| `logprobs=false` | Accepted as a semantic no-op and disclosed through `x-experiential-ignored-parameters`. |
| Unsupported stop, tool, structured-output, or reasoning semantics | Local field-specific `400`. Proven no-ops are removed and disclosed. |
| Unknown top-level fields | Rejected by the closed Chat, Responses, and Messages manifests. |

## Rust gateway

This covers the Rust gateway. Python remains the catalog, authorization, route-resolution, and provider-payload authority. It validates the complete waterfall and freezes exact upstream bytes before native dispatch; Rust sends those bytes without rebuilding or reinterpreting generation parameters.

Vertex now exposes a native `gemini_generate_content` wire profile with the exact model capability flags and publisher-scoped SSE route. A parity test proves its Python payload and Rust-bound frozen payload are identical, including unsupported sampling omission, `top_k`, reasoning translation, and the output-token limit.

The native path also preserves OpenAI reasoning-summary stream events, commits waterfall selection after summary output, emits the Responses summary lifecycle, includes the selected reasoning configuration, and returns the same public errors and ignored-parameter disclosure as Python. Served parity tests prove no upstream dispatch occurs for rejected parameters. Provider-controlled accumulator maps are bounded by both retained bytes and entry count.

## Validation

- Full non-live Python run: 2,745 passed, 2 skipped. Two local process/PTY resource-pressure fixtures each passed immediately in isolation.
- Focused provider-stream suite: 50 passed. Terminal suite: 76 passed.
- Rust: Cargo format and strict clippy pass; 60 no-default-feature unit tests passed.
- Ruff check and format, `ty check`, release-evidence checks, and lock validation pass.
- Exact-head GitHub Actions currently cannot start because the organization account is locked for a billing issue. Jobs contain zero steps and the GitHub annotation identifies the billing lock.

## Release note

This PR moves Experiential to `0.5.7` and the native crate/wheel contract remains `0.1.14`. Platform PR #809 consumes the expanded contract while retaining its released pins until a separately authorized package release and pin update. No merge, publish, or deployment is included here.


kfallah:
@greptile-apps review\n\nPlease review exact head bd4da8ce99f21575ee19f6c7aabe04c99dfac9bd. The prior P1 is addressed by skipping empty summary deltas and enforcing a shared retained-entry ceiling across native reasoning and tool accumulators, with adversarial Python and Rust tests.

kfallah:
@greptile-apps review. Please review exact head 15fe8879d64bd66456c9b7d4477d9c7af49a0434 for provider-wide generation-parameter validation and native Rust frozen-body parity.

kfallah:
@greptile-apps review. Please review exact head 270bb0f2a110005f3c697f8af3aeb2a5d6f0ee9c. The 4/5 P1 is fixed with a 64 MiB cumulative UTF-8 byte ceiling in the Python reasoning-summary parser, atomic pre-mutation admission, and adversarial multibyte repeated-fragment coverage.

kfallah:
@greptile-apps review. Please review exact content-identical CI retrigger head c9fb879bf5734efbeebe8f264222afa860da6b2a.

kfallah:
@greptile-apps review. Please review exact content-identical CI retrigger head 5fac7f6b767f871d8a7f6ef866f23b288373bd47.

kfallah:
@greptile-apps review. Please review exact head fd0133237ccb154746bdaecfc81586321f428c29, including the token-alias regression test and safe CI rerun controls.

kfallah:
@greptile-apps review. Please review exact fix head 106490166e5f553f7055922a44656ad6eabba370. The Python OpenAI Responses, Anthropic, OpenAI-compatible, and Bedrock tool accumulators now share the bounded provider-output byte budget, with provider-level overflow regressions.

kfallah:
@greptile-apps review. Please review exact rebased fix head 733fc8b05019f6a3a368c835b0a539354b945dcf.

kfallah:
@greptile-apps review. Please review exact fix head 50af99b80dd44ee1ca543bfe3d766f8d16cb81eb; the provider stream-state refactor only resolves the repository line-limit gate.

PR #629, opened 2026-08-25
Delete the python gateway data plane

kfallah:
> Previously stacked on #628, which has now merged to `main`; the base is back on `main` and the landed #628 series (including its post-review commits) is merged and reconciled here.

## What this does

Release 0.6.0 removes the python gateway data plane. The compiled native engine (`exp-gateway-native` 0.2.0) is the only engine: every launch path (`exp`, `exp run`, `exp --project`, and the `exp.load_router` compatibility client) serves through it, and a missing compiled extension fails the launch with the exact build command instead of falling back.

## Reconciliation with #628 (parameter gating)

#628's behavior survives the deletion in full, on the surviving native path:

- Parameter gating runs in native admission: `route_generation_parameter_requests` and the capability-aware `preflight_gateway_request` execute in `native_bridge.admit` before any payload is frozen; a `ProviderParameterError` finalizes the accepted request content-free and returns the public field error with no attempt row.
- The landed #628 series' retained provider-entry bounds (`reserve_tool_entry`/`reserve_summary_entry`) survive in the Rust dialects; only the python mirrors of that bounding died with the engine.
- The catalog capability intersection (`_resolved_wire_profile`) moved to `exp/runtime/gateway/execution_resolution.py`, now trimmed to the native branch only; the `python_stream` compatibility dialect died with the python engine. New `execution_resolution_test.py` covers the intersection and the identity fail-closed check.
- Provider clients keep #628's capability-gate constructors, reasoning-effort translation (`reasoning_compat.py`), and expanded wire profiles; only their executor-only `stream()` seams are deleted.
- #628's engine-parity served tests are ported to native-only assertions (`native_messages_test.py`: unsupported reasoning effort, unsupported sampling, catalog generation limits with zero upstream dispatch).

## Reconciliation with #618 (Vertex AI)

Vertex serves natively: `VertexClient` keeps #628's capability-gated constructor and resolves through the shared `gemini_generate_content` dialect wire profile with the publisher-scoped SSE endpoint and a current OAuth bearer token (minted on the bridge's blocking-callback thread, never an event loop). Its executor-only `stream()` seam is deleted with the engine.

## Regenerated goldens

Commit 1 froze the encoder parity contract as committed golden bytes; #628 then changed the encoders. The goldens in `exp/runtime/gateway/testdata/parity_goldens.json` are regenerated from the post-#628 python encoders, so the frozen contract now includes the Responses reasoning-summary lifecycle (`rs_` items, `reasoning_summary_part/text` events, selected reasoning configuration in bodies). Two contracts are newly frozen: the Chat `x-experiential-ignored-parameters` disclosure (new golden plus a fixture argument on `encode_chat_fixture`) and the Messages surface dropping reasoning-summary deltas without changing its bytes. Every Rust encoder passes byte-for-byte against the new goldens, and against the live python encoders where those survive.

## Differential loadtest findings (2.2M requests, report in comments)

- Fixed: a successful attempt terminating with zero semantic output (empty completion, or max_tokens truncating before any text) settled with the collected usage but dropped the usage event from the outward stream, so callers lost the token accounting on all six surfaces while billing stayed correct. The waterfall now returns the tracked usage ahead of the terminal; twelve served-engine repros (Chat/Responses/Messages, streaming and not, empty and truncated) were written failing first.
- Fixed: streaming responses again serve `content-type: text/event-stream; charset=utf-8`, matching the prior engine for strict clients.
- Named follow-up (bounded, non-failing): ledger calls run via `spawn_blocking` + `Python::attach`, and `persistent_connection` caches one SQLite connection per blocking thread in a `threading.local`. Tokio blocking threads retire without closing the cached connection, so `gateway.db` descriptors track ever-seen blocking threads (plateaued at 350 FDs / ~300 MB RSS in a 608k-request soak; bounded by the pool's thread ceiling, nothing fails). Candidate fixes: a dedicated small blocking-thread pool for bridge calls, a checkout/return connection pool instead of thread-locals, or thread-exit cleanup that closes the cached connection. Deferred because it changes the ledger connection strategy across every bridge call site, which deserves its own soak-validated change.

## Deleted

- `GatewayService` request serving, `create_gateway_app`, and the uvicorn engine path.
- `GatewayExecutor` and the execution machinery (`execution.py`, `aggregation.py`, `stream_attempts.py`), plus the per-provider `stream()` seam and the python event mappers (`providers/streaming.py`, `streaming_events.py`, `gemini_streaming.py`, `bedrock_streaming.py`, `streaming_usage.py`) whose only consumer was that executor.
- The python Anthropic Messages encoders and error envelope (`anthropic_protocol/encoding.py`, `errors.py`); the native plane renders that surface.
- The public embeddable composition seam (`create_gateway_runtime`, `GatewayRuntime`, `GatewayRuntimeConfig`) and its docs section.
- The `--engine` flag (passing it is a usage error) and the deprecated `fallback_app`/`fallback_port` relay in the crate.
- The `fastapi` and `uvicorn` dependencies.
- Every test that existed solely for the deleted engine; the rest were converted to the component seam (`load_gateway_components`) or to the served native engine on real sockets.

## Kept

Authority/store, ledger and group-commit, budgets, routing, `DeploymentHealthRegistry` (bridge-consumed), catalog lifecycle and hot reloader, all protocol decoders and payload builders, `boundary.py` error mapping, the whole `native_*` bridge stack, usage report and HTML via native callbacks, guardrails, replay/continuation stores, and the CLI native serving path.

## Behavior notes

- Every encoder parity contract asserts against committed golden bytes (`testdata/parity_goldens.json`); no test imports a deleted python encoder or mapper.
- Readiness: the accounting latch lives on the components (`accounting_healthy`), so hosted compositions report their own store health and the bridge ANDs it with its settlement latch. Nothing references `executor.require_healthy` or `mark_accounting_unhealthy`.
- Startup keeps the credential/route proof: loading granted aliases builds each alias's credential-free route proof with actionable per-alias reasons, and native servability of every pool deployment is validated before binding.
- An `{"escalate": reason}` admission disposition is classified for content-free metrics and fails closed with the shared internal error; nothing hands a request to another engine. The accepted request is finalized content-free (no attempt row) before the disposition returns.
- Two regressions the installed-wheel release evidence caught during this work are fixed with unit repros: a continued Responses request now joins its first turn's selection episode (no second request-time embed), and ledger acceptance runs before route selection so a keyed replay that must fail closed never triggers learned-selection embedding.

## Versions

Flagship `experiential` 0.6.0; `exp-gateway-native` 0.2.0 with the flagship floor `>=0.2.0,<0.3`; the lockstep test derives the ceiling from the crate version. `cargo update -p exp-gateway-native` and `uv lock` are current. (main carries 0.5.6/0.1.14 from #628; this PR supersedes both with the strictly greater pair.)

## Accepted boundary

Platforms without a prebuilt wheel (musl, exotic architectures) must install from sdist and need a Rust toolchain to build `exp-gateway-native`. There is no interpreted fallback.

## Release notes draft (0.6.0)

- The gateway now has exactly one data plane: the compiled native engine serves Chat Completions, Responses, Anthropic Messages, models, health, and usage on every launch path.
- Removed: the python serving engine, the `--engine` flag, and the embeddable `create_gateway_runtime` composition API. `exp.load_router` still returns an official OpenAI client, now backed by a private native gateway with a graceful stop handle.
- Removed dependencies: `fastapi`, `uvicorn`.
- Breaking: installs on platforms without a prebuilt `exp-gateway-native` wheel require a Rust toolchain (sdist build). A missing extension fails launch with the exact build command.


kfallah:
@greptile-apps review

kfallah:
@greptile-apps review

kfallah:
@greptile-apps review

kfallah:
@devin can you do in-depth integration testing with high volume traffic to ensure no regressions? spend up to $1k

kfallah:
@greptile-apps review

kfallah:
@greptile-apps review

kfallah:
Thanks for the differential run — the no-output diagnosis was exact. Status on the three items, fixed on this branch:

**1. Zero-output usage drop: fixed.** The cause was where you pointed: the waterfall's "successful terminal with no semantic output" branch settled with the collected usage and then returned only the terminal event, so no encoder ever saw a usage event. The branch now returns the tracked usage event ahead of the terminal. Twelve served-native repros were written failing first and now pass: Chat, Responses, and Messages, streaming and non-streaming, for both the empty completion and the max_tokens truncation, against a mock upstream that returns zero content plus a real usage payload (`exp/runtime/gateway/tests/native_messages_test.py::test_*_zero_output_*`). Chat non-stream returns `prompt_tokens` with `completion_tokens: 0`, Chat streaming emits the `include_usage` final chunk, Responses carries usage on both the completed and incomplete terminals, and Messages reports the real input count.

**2. Bounded FD/connection retention: deferred as a named follow-up**, written up in the PR body. Your mechanism read matches the code: `spawn_blocking` + `Python::attach` with `persistent_connection` caching one SQLite connection per blocking thread in a `threading.local`, and retiring blocking threads never close it. The candidate fixes (dedicated small blocking pool, checkout/return connection pool, or thread-exit cleanup) all change the ledger connection strategy across every bridge call site, so it gets its own soak-validated change rather than riding this deletion. It is bounded by the pool ceiling and caused no failures in your 608k soak.

**3. SSE content-type: restored.** Streaming responses serve `text/event-stream; charset=utf-8` again on every surface, including replayed bodies.

[Posted by the PR branch's agent]

kfallah:
@greptile-apps review

PR #635, opened 2026-08-28
Health-aware routing: fail-fast timeouts, skip dead lanes, admit past disabled rungs

SilenNaihin:
## Health-aware routing for the native gateway

Production evidence (live, this session): models fall over ~35s per dead
deployment rung because there is no fast timeout; the per-pod in-memory circuit
breaker doesn't share state, so fresh pods keep re-trying dead lanes; and
admission throws on a disabled lead rung instead of skipping it. Concretely,
`qwen3.8-27b` 35s-times-out on both its lanes, and 187 models have a default
waterfall that leads with a disabled deployment.

Three independent, additive fixes, each behind a clear boundary so review can
accept them separately. The gateway data plane is Rust; candidate selection and
admission stay in the Python control plane.

### 1. Fail-fast timeouts on the upstream call
The native client hard-coded a 10s connect timeout and used the deployment's
own (~35s) timeout as the bound for both the open phase and every chunk read,
so a lane that accepted the connection but never streamed a byte held the
request for the full ~35s before failing over.

- New `ServeConfig` knobs `connect_timeout_seconds` (**default 5.0**) and
  `time_to_first_byte_seconds` (**default 15.0**), plumbed through
  `serve_native_gateway`.
- `build_client` now takes the connect timeout (`upstream.rs`).
- `UpstreamRelay` owns a per-attempt first-byte deadline; `next_event` bounds
  the wait for the first byte by it and, on a genuine first-byte stall,
  classifies a `Timeout` (`relay.rs`). `run_attempt` bounds the open phase by
  the same window (`waterfall.rs`).
- **Does not cap generation**: once the first byte arrives, reads revert to the
  deployment's per-chunk timeout, so slow reasoning models stream as long as
  they need.

Before: a stalled lane blocks ~35s. After: it fails over in ~≤15s (first-byte)
or ~5s (connect). Test: `relay::tests::a_stalled_first_byte_trips_the_ttft_bound_not_the_chunk_timeout`.

### 2. Skip a dead lane instead of redialing it (health memory)
A first-byte stall is now **failover-eligible but not same-deployment
retryable** (`relay.rs::first_byte_timeout_failure`), so the waterfall advances
straight to the next live rung rather than wasting a redial on a lane that
never answered. Combined with fix 1, a fresh pod's cost on a dead lane is ~one
fail-fast window instead of several.

Design choice — option (b) with a documented follow-up. Option (a), a
fleet-shared TTL'd health signal read at route time, is the real cross-pod fix,
but the OSS engine's store is per-pod SQLite, so a true fleet-shared signal
needs a shared store and is a fragile distributed-state change to force into one
PR. The per-pod circuit is left at its **two-failure default** deliberately:
lowering it to one failure breaks the frozen bounded same-deployment redial (a
single transient blip should still redial), removes transient retries for
singleton pools, and destabilizes the module-scoped waterfall scenario suite in
an order-dependent way. The fleet-shared design is documented as the follow-up
in `health.py`, layering onto the same `claim`/`failed` seam. Test:
`test_default_threshold_preserves_one_transient_redial_before_opening`.

### 3. Admission skips disabled/unavailable rungs
`resolve_route_profiles` resolved the whole route all-or-nothing, so one
disabled lead rung raised out of the per-deployment wire build and surfaced as
the mislabeled `INTERNAL` "gateway admission failed before provider dispatch".

- Replaced with `resolve_dispatchable_wires` (`native_execution.py`): resolves
  each rung independently and keeps only the dispatchable ones, in route order.
  A disabled/drifted connection (`ModelConnectionError`), an identity mismatch,
  or a non-dialect capability rejection skips that rung; a request-scoped fault
  still fails the whole admission when the surviving wires are built.
- `route_with_dispatchable_deployments` narrows the executable route to the
  survivors; per-rung health keys, idempotency keys, and route-depth headers
  stay aligned (they are per-deployment, not order-dependent).
- If **every** rung is unavailable for a non-dialect reason, admission finishes
  the accepted request with an honest `PROVIDER_INTERNAL` terminal ("all
  certified deployments for this model are currently unavailable") — not a hang,
  not a mislabeled internal. A route with **no native dialect anywhere** still
  escalates, preserving the launch-time servability contract. The keyed-replay
  servability probe uses the same resolver.
- Removed the now-unused `resolve_route_profiles` and updated its doc
  reference.

Before: a disabled position-0 deployment fails the whole request. After: it is
skipped to the next live rung. Tests:
`test_a_disabled_lead_rung_resolves_to_the_next_live_rung`,
`test_every_rung_disabled_raises_the_unavailable_terminal`,
`test_every_rung_dialectless_preserves_the_escalation_contract`.

## Validation
- Rust: `cargo test --no-default-features` — 66 passed; `cargo clippy` clean;
  `cargo fmt --check` clean.
- Python: full `exp/runtime/gateway` suite — 475 passed; router + CLI gateway
  callers of the changed `serve_native_gateway` — 128 passed.
- `ruff check .`, `ruff format --check`, and `ty check exp/runtime exp/cli/gateway`
  all pass.

### DIFF
--- a/exp/runtime/gateway/health.py
+++ b/exp/runtime/gateway/health.py
@@ -1,4 +1,24 @@
-"""Bounded per-deployment circuit and throttle state for exact-model waterfalls."""
+"""Bounded per-deployment circuit and throttle state for exact-model waterfalls.
+
+This registry is per-process: each gateway pod keeps its own circuit state, so a
+lane that one pod has proven dead is still probed by every other pod. Two things
+keep that probe cheap rather than a 35-second stall: the data plane's fail-fast
+connect and time-to-first-byte bounds detect the dead lane in seconds, and a
+first-byte stall is classified failover-eligible but not same-deployment
+retryable, so the pod advances to the next live rung instead of redialing the
+stalled lane.
+
+Follow-up (fleet-shared health): to route the *whole* fleet around a dead lane
+without every pod paying its own probe, this in-memory state should be backed by
+a small shared, TTL'd health signal the control plane reads at route time and
+writes on failure (keyed by the same ``DeploymentHealthKey`` tuple). That
+cross-pod state change is deliberately kept out of this change: the OSS engine's
+store is per-pod SQLite, so a true fleet-shared signal needs a shared store and
+is a fragile distributed-state change to force into one PR. It layers onto the
+same ``claim``/``failed`` seam without changing the waterfall contract, so the
+per-pod fast-fail-and-skip above is the conservative first step and the shared
+signal is the follow-up.
+"""
 
 from __future__ import annotations
 
@@ -47,6 +67,13 @@ class DeploymentHealthRegistry:
         clock: Callable[[], float] = time.monotonic,
     ) -> None:
         """Initialize finite in-process deployment health policy.
+
+        The default opens a circuit after two consecutive operational failures,
+        which preserves the bounded same-deployment redial: a single transient
+        blip still redials the lane, and only a lane that fails twice is
+        suppressed for the cooldown. A stalled (dead) lane is skipped before it
+        reaches this threshold because a first-byte timeout is classified
+        failover-not-redial in the data plane.
 
         Args:
             failure_threshold: Consecutive operational failures that open one circuit.
--- a/exp/runtime/gateway/health_test.py
+++ b/exp/runtime/gateway/health_test.py
@@ -68,6 +68,36 @@ def test_operational_failures_still_open_the_circuit() -> None:
         assert not registry.claim(_KEY)
 
 
+def test_default_threshold_preserves_one_transient_redial_before_opening() -> None:
+    """The default keeps a lane admissible after a single transient failure.
+
+    A stalled lane is skipped by the data plane's failover-not-redial timeout
+    classification, so the circuit deliberately tolerates one blip here rather
+    than suppressing a lane that a bounded redial could still recover.
+    """
+    registry = DeploymentHealthRegistry(clock=lambda: 100.0)
+
+    registry.failed(_KEY, _failure(GatewayFailureClass.PROVIDER_INTERNAL))
+    assert registry.claim(_KEY)
+
+    registry.failed(_KEY, _failure(GatewayFailureClass.PROVIDER_INTERNAL))
+    assert not registry.claim(_KEY)
+
+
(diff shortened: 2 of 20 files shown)


SilenNaihin:
## W16 evidence failures: root cause and fix

**Root cause — stale lockfile, not an evidence-shape regression.** This PR bumped `exp-gateway-native` 0.2.1 → 0.2.2 in both `pyproject.toml` (dependency constraint `>=0.2.2`) and `exp/runtime/gateway/native/pyproject.toml` (package version), but `uv.lock` was left pinned at 0.2.1. The `w16-darwin-evidence` job's install step runs `uv sync`, which sees the lock is out of date and re-resolves it, rewriting `uv.lock` in place and dirtying the working tree. Both W16 tests open with `exact_checkout_revision()`, whose `git diff --quiet` release-evidence guard then fails:

```
RuntimeError: release evidence requires a checkout with no tracked source changes
```

That single guard is why *both* tests fail identically (`FF`). The health-aware routing serving/evidence output shape is unaffected — the evidence assertions pass once the checkout is clean.

**Fix.** Bump the locked `exp-gateway-native` version to 0.2.2 so the committed lock already satisfies `pyproject.toml`. `uv sync --extra dev` is now a no-op (`Resolved … Checked …`, no `Updated`), so the checkout stays clean and the guard passes. One-line change to `uv.lock`; no test assertion was weakened and no production code changed.

**Verification (local, macOS/arm64, Python 3.12):**
- Two W16 evidence tests on the committed (clean) tree: **2 passed**.
- `uv sync --extra dev` on the fixed lock: no relock, tree stays clean.
- `ruff format --check`: 694 files already formatted. `ruff check`: All checks passed. `ty check`: All checks passed.

Commit: `5f3c1f1a`.

SilenNaihin:
Closing as superseded. Main independently shipped the core of this PR: per-deployment circuit breakers + throttle + failover live in `exp/runtime/gateway/health.py` (skip dead lanes, open a circuit after consecutive failures, recovery probes, admit-past-open-circuit). The one genuinely-missing residual — fail-fast connect/time-to-first-byte timeouts that trip that existing circuit — is extracted clean in #640. The only remaining unmerged behavior is skipping a *statically* disabled/drifted lead rung at admission, which is an admission-semantics change that should be re-proposed on main's `select_route_deployments` idiom (not the replaced `route_with_dispatchable_deployments`) if we want it.

PR #637, opened 2026-08-28
Add TrustedRouter as a native OpenAI-compatible provider

jperla:
[TrustedRouter](https://trustedrouter.com) is an attested OpenAI-compatible router. It serves the standard Chat Completions surface from a fixed origin, so it fits the native provider path rather than requiring every operator to hand-declare an `openai-compatible` connection and its capabilities.

```bash
exp config providers --provider trustedrouter   # TRUSTEDROUTER_API_KEY
```

It routes through the existing `OpenAICompatibleClient` — no new adapter.

### Listing

The catalog at `https://api.trustedrouter.com/v1/models` is OpenRouter-shaped and names its parameters with the same vocabulary, so `supported_parameters` is read the way OpenRouter's is: tools, structured outputs, sampling bounds, logprobs, and reasoning.

That array is treated as **optional**. A model that omits it keeps those capabilities *unknown* rather than reported as unsupported, and the operator declares the minimum fields for a selected role exactly as with an identity-only OpenAI-compatible model. This matters because the currently published catalog omits it on every entry — reusing `_openrouter_model` unchanged would have read the missing array and silently marked all 638 models as supporting no tools, no sampling, and no reasoning. The roles, context window, and prices are published today and are always kept.

I ran the live catalog body through the parser. All 638 entries parse — 574 chat, 28 embedding, 638 priced, 596 with a context window; the other 42 report `context_length: 0`, which `_positive_int` already turns into an unknown window rather than a bogus one. There are tests for the published-array case, the absent-array case, and the zero-context case.

### Certification

`trustedrouter` uses `_compatible_provider_cells`, the same honest shape as `azure` and `openrouter`: non-live capabilities are `INHERITED_COMPATIBLE_FIXTURE_PASS`, since it genuinely runs on the generic compatible adapter whose fixtures those cells cite, and `CREDENTIAL_GATED_LIVE` stays `NOT_RUN_REQUIRES_CREDENTIALS`. No live run is claimed. I left `_EVALUATED_AT` at its existing date since the inherited evidence is unchanged — say the word if you would rather re-date the artifact.

### Test updates

Adding a provider row shifts the setup picker, so the positional walks in `provider_picker_test.py`, `cli/gateway/setup_test.py`, and the installed-wheel walk in `release_test.py` move by one. Those are index updates only; each test still selects the same provider it did before.

`just gate` is green: `ruff check`, `ruff format --check`, `ty check`, and `pytest` (2618 passed, 2 skipped).

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -23,6 +23,9 @@ OPENAI_API_KEY=
 # OpenRouter (serves pool candidates like GLM-5.2 / Kimi, priced from its own catalog)
 OPENROUTER_API_KEY=
 
+# TrustedRouter (attested OpenAI-compatible router; catalog publishes roles, context, prices)
+TRUSTEDROUTER_API_KEY=
+
 # Tinker (distillation: student sampling, training, and serving trained adapters)
 TINKER_API_KEY=
 
--- a/docs/reference/providers.md
+++ b/docs/reference/providers.md
@@ -8,7 +8,7 @@ Configure connections with `exp config providers` or the first `exp build` on a 
 An interactive terminal opens a provider list: Up and Down move focus, Enter selects or deselects
 the focused provider, and the Complete row submits the selection. Agents skip that list with
 repeatable `--provider` flags (`experiential-cloud`, `openai`, `anthropic`, `gemini`, `openrouter`,
-`openai-compatible`, `azure`, `bedrock`). Unsupported or duplicate values
+`trustedrouter`, `openai-compatible`, `azure`, `bedrock`). Unsupported or duplicate values
 fail before any catalog write. Azure and Bedrock still require manual model IDs. Other selected
 providers use account model discovery when credentials are available.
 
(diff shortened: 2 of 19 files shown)


jperla:
Thanks for taking a look. Could you say what led to the close? I want to make sure I address the actual reason rather than guess at it.

I've opened #638 with the review finding fixed: TrustedRouter models with unknown capabilities were dropped in `prepare_providers` and rejected by `can_declare_role`, so they could not be given a judge or router-candidate role. Those were two spellings of one rule, so #638 gives it a single owner (`OPERATOR_DECLARED_PROVIDERS`) that `declaration.py` reads, and adds `trustedrouter` to it.

If the close was about something else — provider scope, the certification artifact, or not wanting fixed-origin routers beyond OpenRouter — I'd rather know than keep iterating on the wrong thing, and I'm happy to close #638 too.

PR #638, opened 2026-08-28
Add TrustedRouter as a native OpenAI-compatible provider

jperla:
Reopens #637 with the review finding fixed. [TrustedRouter](https://trustedrouter.com) is an attested OpenAI-compatible router; it serves the standard Chat Completions surface from a fixed origin, so it fits the native provider path rather than requiring every operator to hand-declare an `openai-compatible` connection and its capabilities.

```bash
exp config providers --provider trustedrouter   # TRUSTEDROUTER_API_KEY
```

Execution routes through the existing `OpenAICompatibleClient` — no new adapter.

### What changed since #637

The review on that PR was right: a TrustedRouter model whose capabilities were unknown could not be assigned a judge or router-candidate role, and there was no way to declare the missing metadata. Two gates caused it — `prepare_providers` dropped unknown rows for any provider outside `_OPERATOR_DECLARED_PROVIDERS`, and `can_declare_role` separately hardcoded `"openai-compatible"`.

Those were two spellings of one rule, so this gives it a single owner, `OPERATOR_DECLARED_PROVIDERS`, which `declaration.py` now reads instead of its own string literal. `trustedrouter` joins it, so an unproven TrustedRouter model stays visible and the operator declares the minimum fields the selected role needs — the same flow an identity-only OpenAI-compatible model already gets. Covered by a test that fails against the old gate.

### Listing

The catalog is OpenRouter-shaped and names its parameters with the same vocabulary, so `supported_parameters` is read the way OpenRouter's is: tools, structured outputs, sampling bounds, logprobs, and reasoning.

The array is treated as **optional**. A model that omits it keeps those capabilities *unknown* rather than reported as unsupported. That matters because the currently published catalog omits it on every entry — reusing `_openrouter_model` unchanged would have read the missing array and silently marked all 638 models as supporting no tools, no sampling, and no reasoning. Roles, context window, and prices are published today and are always kept.

I ran the live catalog body through the parser. All 638 entries parse — 574 chat, 28 embedding, 638 priced, 596 with a context window; the other 42 report `context_length: 0`, which `_positive_int` already turns into an unknown window rather than a bogus one. There are tests for the published-array case, the absent-array case, and the zero-context case.

### Certification

`trustedrouter` uses `_compatible_provider_cells`, the same shape as `azure` and `openrouter`: non-live capabilities are `INHERITED_COMPATIBLE_FIXTURE_PASS`, since it genuinely runs on the generic compatible adapter whose fixtures those cells cite, and `CREDENTIAL_GATED_LIVE` stays `NOT_RUN_REQUIRES_CREDENTIALS`. No live run is claimed. I left `_EVALUATED_AT` at its existing date since the inherited evidence is unchanged — happy to re-date the artifact if you would rather.

### Test updates

Adding a provider row shifts the setup picker, so the positional walks in `provider_picker_test.py`, `cli/gateway/setup_test.py`, and the installed-wheel walk in `release_test.py` move by one. Those are index updates only; each still selects the same provider it did before.

`just gate` is green: `ruff check`, `ruff format --check`, `ty check`, and `pytest` (2620 passed, 2 skipped).

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -23,6 +23,9 @@ OPENAI_API_KEY=
 # OpenRouter (serves pool candidates like GLM-5.2 / Kimi, priced from its own catalog)
 OPENROUTER_API_KEY=
 
+# TrustedRouter (attested OpenAI-compatible router; catalog publishes roles, context, prices)
+TRUSTEDROUTER_API_KEY=
+
 # Tinker (distillation: student sampling, training, and serving trained adapters)
 TINKER_API_KEY=
 
--- a/docs/reference/providers.md
+++ b/docs/reference/providers.md
@@ -77,10 +77,11 @@ with the same vocabulary. Setup keeps the completion and embedding roles from
 
 `supported_parameters` is read the way OpenRouter's is when a model publishes it. The array is
 optional: a model that omits it keeps tool, structured-output, sampling, and reasoning support
-**unknown** rather than reporting them as unsupported, and the operator declares the minimum
-fields for a selected role the same way an identity-only OpenAI-compatible model does. A model
-that reports a non-positive `context_length` keeps an unknown context window rather than borrowing
-one from a neighboring field.
+**unknown** rather than reporting them as unsupported. Because that listing can return an unproven
+model, `trustedrouter` joins `openai-compatible` in the operator-declaration path, so such a model
+stays visible and the operator declares the minimum fields a selected role needs instead of the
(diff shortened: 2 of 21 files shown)


kfallah:
@jperla hey, this can be added by users via the openai compatible endpoint in the provider config. we are not planning to support TrustedRouter right now.

jperla:
can you explain? OpenRouter is right there, and it's not end to end encrypted, and it's no work i did all the work

PR #665, opened 2026-08-29
gateway: unblock Codex reasoning echoes and Claude Code context_management (0.7.5)

SilenNaihin:
## What (the two agent-activation blockers, lean 0.7.5 first slice)

**1. Codex turn-2 reasoning echo (Responses, store=false).** Root-caused live with variant isolation: OpenAI cryptographically binds `encrypted_content` to its ORIGINAL reasoning item id and rejects any mismatch ("Encrypted content item_id did not match"). The gateway mints its own stable public item ids, Codex echoes those, and the builder forwarded the echoed id upstream, so every continuation with a reasoning item 400ed. Verified matrix: verbatim-original-id passes, id-less passes, foreign id fails by name. Fix: the replayed reasoning item's `id` is NEVER forwarded upstream; an id-less item verifies against the id embedded in the encrypted payload itself. Callers keep sending ids; they stay accepted and simply never cross. Answering the triage questions: it was NOT an engine gate or a catalog capability - the request reached OpenAI and the 400 was OpenAI's, surfaced through the sanitized provider-transport classification ("provider rejected the request..."), which is why no capability was named; that message is deliberately content-free because provider 400 bodies are not relayed. No catalog change needed. Live proof: the exact failing shape (gateway-minted-id echo) re-driven through the fixed builder completes the two-turn loop (200, response.completed).

**2. Claude Code `context_management` (Messages).** Anthropic's native context-editing config, sent by default by the real CLI, was one of the conscious UNSUPPORTED classifications. Upgraded with exactly the 0.7.4 `cache_control` treatment plus one discovery: the live API rejects the field WITHOUT its beta header ("Extra inputs are not permitted"), so forwarding verbatim requires `anthropic-beta: context-management-2025-06-27` (verified live; merged into any operator-declared beta list, injected per request through the wire entry so only requests carrying the field send it). Decode is deliberately SHALLOW (an object, carried verbatim): the nested shape (`edits`/`trigger`/`keep`) is an evolving provider beta and a closed model here would recreate the reject-what-real-clients-send incident class. Digest handling matches the other Anthropic carriers: excluded from serialization (config-free digests byte-identical, pinned by test) with present values folded into `canonical_request_sha256`. Non-Anthropic routes validate-and-drop with `ignored_parameters` disclosure. Classification updated (Anthropic manifest: CONDITIONALLY_SUPPORTED, capability `context_management`); the Responses-surface `context_management` (a different OpenAI feature) stays consciously rejected. Live proof: the real CLI shape through our builders + headers streams clean on claude-fable-5.

**Conscious-rejection audit** (`reasoning.mode`, `moderation`, `prompt_cache_options`, `prompt_cache_retention`; Anthropic `container`/`mcp_servers`/`output_config`/`service_tier`): no evidence any real first-party CLI sends them - Codex's verified request shape is store/include/reasoning{effort,summary}/prompt_cache_key, and Codex session rollouts do not retain raw request bodies (the local `"mode":"default"` hits are turn metadata, not request fields). All stay rejected. Residual risk noted: a proxied capture of one real Claude Code and one real Codex session against staging would upgrade this from absence-of-evidence to evidence, recommended as a platform-side follow-up.

**Structural:** the change pushed `streaming_requests.py` over the 999-line budget; the route-wide generation-control policy moved verbatim to `exp/runtime/models/providers/route_admission.py` (the payload builders keep their module; `fireworks_continuation_required` is shared and now public).

## Ship plan

Lean 0.7.5 first slice: this PR only. The capability-preservation policy (#664, settled clean) trails as 0.7.6 and will need a trivial rebase over the `route_admission.py` split (its coercion seam imports the moved function).

## Gates

`uv run pytest -q`: 2717 passed, 2 skipped, 0 failed. `ruff check .`, `ruff format --check .`, `ty check`: clean. Python-only; crate untouched.

### DIFF
--- a/docs/reference/gateway-architecture.md
+++ b/docs/reference/gateway-architecture.md
@@ -182,7 +182,10 @@ are accumulated in original order and validated only at the complete-call bounda
 authenticate with `x-api-key` (the Anthropic SDK default) or a standard Bearer header; both carry
 the same virtual key, and every failure on this surface is rendered in the Anthropic error
 envelope `{"type": "error", "error": {...}}`. The decoder translates text, `tool_use`,
-`tool_result`, `thinking`, and `redacted_thinking` blocks faithfully (thinking history rides an
+`tool_result`, `thinking`, and `redacted_thinking` blocks faithfully, and carries the caller's
+`context_management` object verbatim (shallow-validated as an object; Anthropic rungs receive it
+byte-for-byte together with its required `anthropic-beta` token, while non-Anthropic routes drop
+it with `ignored_parameters` disclosure) (thinking history rides an
 opaque provider-reasoning carrier with byte-exact signatures, and a caller `thinking`
 configuration is forwarded verbatim on models that honor it, overriding the catalog's
 adaptive default; on the adaptive-only generation, which rejects `enabled`/`disabled`
@@ -242,7 +245,9 @@ continuation and duplicate replay retain content only in bounded, process-local,
 alias-revision-scoped stores. A `store: false` request skips continuation retention entirely
 (continuing from its ID answers `continuation_unavailable`), and
 `include: ["reasoning.encrypted_content"]` forwards the encrypted reasoning request to native
-OpenAI Responses routes, whose opaque payloads replay verbatim from the caller's input. Replay is opt-in through an idempotency or client request key. Restart
+OpenAI Responses routes, whose opaque payloads replay verbatim from the caller's input; the
+replayed reasoning item's `id` is never forwarded upstream because the provider binds the
+encrypted payload to its original item id and callers echo this gateway's own minted public ids. Replay is opt-in through an idempotency or client request key. Restart
 or eviction returns an explicit unavailable error and never reconstructs content from SQLite.
 
 ## Content-free observability and lifecycle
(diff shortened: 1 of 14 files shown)


SilenNaihin:
Reconcile-rebased onto current main (head 35c3f10f) over #653–#657 and #664, per the semantic checklist. Full suite 2900 passed / 0 failed (the 2 release-evidence 'failures' were the clean-tree requirement against uncommitted work; green after commit). Both fixes re-proven LIVE through the rebased builders.

**#655 vs the id-binding fix — reconciled to one behavior: the reasoning item id never crosses upstream.** #655's normalizer now binds and preserves provider item identity downstream (complementary, not contradictory), but its replay tests pinned forwarding the echoed id upstream. The merged contract strips it: an id-less item verifies against the id embedded in encrypted_content itself, so replay is correct regardless of whether the caller echoed a provider-original or gateway-minted id. All ten #655 replay tests now pin the merged semantics.

**New live finding while re-proving the merged shape:** OpenAI rejects `status` on replayed *reasoning* input items outright ("Unknown parameter: 'input[N].status'") while accepting it on function_call and message items — verified by per-item-type isolation. #655's continuation replay forwarded status on reasoning items and would have 400ed the tool-continuation path in production. The merged contract: replayed reasoning items carry ONLY type/summary/encrypted_content; function_call and message items keep id and status exactly as #655 shipped them. Live proof through these builders: two-turn Codex-shaped loop with a foreign echoed id completes (HTTP 200).

**#657 vs encrypted content on Fireworks rungs:** no collision — #657's sealed carrier is the gateway-internal OpaqueReasoningContentBlock on the Fireworks lane; EncryptedReasoningBlock remains the OpenAI lane, and the homogeneity gates are untouched.

**#654 / #664 module layout:** this branch's route_admission.py split is DROPPED — main solved the same 999-line budget via generation_parameter_validation.py (#657), and #664's native_admission.py already imports route_generation_parameter_requests from streaming_requests. The context_management payload emission and disclosure now live in main's existing structure (streaming_requests.py, 999-budget clear at ~1000 lines... 986+14).

**context_management re-proof:** builders + anthropic-beta context-management-2025-06-27 header → HTTP 200 end_turn on claude-fable-5.

PR #689, opened 2026-08-31
anthropic: serve server tools end to end (web_search) with verbatim block carry

SilenNaihin:
## Problem

Live repro on production 0.7.10 (2026-08-31): a real Claude Code session invoking WebSearch got

```
400 "Invalid value for 'tools.0.input_schema': Field required"
```

Server-tool definitions (`web_search_20250305`-style typed entries) carry no `input_schema`, and the strict tool wire model required one. The #685 drift gate classified custom-tool fields but not the tool-type union. Worse, acceptance alone would not have fixed the feature: the native normalizer skipped unknown `server_tool_use` block starts and then failed the stream at their first `input_json_delta` ("provider emitted arguments before a tool start"), result blocks would have been silently lost, echoed history blocks were hard-rejected on turn 2, `pause_turn` was rewritten to a fake `end_turn`, and search-inflated input tokens were under-billed. This PR lands the lane end to end.

## Live verification (api.anthropic.com, house key, 2026-08-31)

- Every GA `ToolUnionParam` type probed bare on a plain key: **all accepted with no beta header**; the only 400s are the provider's own per-model support errors ("'claude-sonnet-4-6' does not support tool types: ..."), so the provider stays the authority per model and no beta injection is needed anywhere.
- A full live `web_search` stream was captured and is pinned (neutralized) as a golden normalizer fixture. Real-wire facts it encodes: `server_tool_use` opens with empty input then streams `input_json_delta` fragments (the first one empty); `web_search_tool_result` arrives whole in its start frame with a `caller` object; the answer text block opens with a `citations` key and a leading `citations_delta`; and `message_delta.usage` carries the search-inflated cumulative input total (2230 -> 10538) that `message_start` does not.
- End-to-end proof: the exact production repro body decoded through the new path, the gateway-built payload streamed a real search live (200), the native normalizer produced two verbatim `server_tool_block` events + text + correct usage, the aggregated assistant content was echoed back through decode as turn 2, and the provider accepted the replay and answered from the prior search result.

## Decisions

**Accepted (forward verbatim on Anthropic rungs, at exact caller positions):** all 22 GA union types (web_search/web_fetch/code_execution/tool_search families, bash, text_editor, memory, browser/computer toolsets). Decode validates shallowly against the recorded type table (`MESSAGES_SERVER_TOOL_TYPES_ACCEPTED`); the raw entry rides a new serialization-excluded `GatewayRequest.provider_server_tools` carrier that joins replay identity when present. Server tool names join the shared tool-name space (collisions reject; a named `tool_choice` may select one).

**Echoed history:** `server_tool_use` + the six `*_tool_result` block types decode onto a per-block opaque carrier (`GatewayMessage.provider_server_tool_block`, assistant-only, sole content, mirroring `provider_native_item`) and re-emit byte-for-byte in order; the emitter merges consecutive assistant messages back into one content list.

**Non-Anthropic rungs: NAMED REJECT, not disclosure-drop** — for both definitions and history blocks. Rationale (recorded in code): a server tool changes what the model can do; silently dropping a search capability the caller declared is a behavior lie. This matches the `thinking`/`tool_is_error` precedent for Anthropic-only semantics.

**Recorded rejections with rationales:** `mcp_toolset` (binds the rejected `mcp_servers` field), `advisor_20260301` (a second, differently priced model inside one request falsifies committed route identity/billing — the fallbacks rationale), legacy `bash_20241022`/`text_editor_20241022` and `computer_2024*/2025*` (beta-gated, unverified). Content blocks `search_result` (gateway serves no citations) and `container_upload` (rejected `container` field) get named hints.

## Response side (Rust, crate 0.3.11 -> 0.3.12)

- Normalizer: a closed `ANTHROPIC_SERVER_TOOL_BLOCK_TYPES` set accumulates opaque blocks (verbatim start frame + folded `input_json_delta` fragments, byte/entry budgets enforced) and emits one `ServerToolBlock` event per block at its stop frame; genuinely unknown block kinds keep today's skip.
- Messages encoder: server blocks re-emit whole in their `content_block_start` (the `redacted_thinking` precedent), never flip the `tool_use` stop reason, and aggregate verbatim into the non-streaming body.
- `pause_turn` is now a first-class `Paused` terminal: the Messages surface reports the provider's own `pause_turn` stop reason (streaming and aggregated), settlement bills it like completion, and the Chat/Responses surfaces (which can never legally see it) map it defensively to completion.
- Guardrail redaction drops server-tool blocks so rewritten output cannot leak searched/fetched content; a `Paused` terminal still receives the replacement text.
- **Billing fix:** `message_delta` usage counts now supersede `message_start` when present, so searched content injected mid-turn bills at the provider's cumulative total instead of the pre-search count (live turn: 2230 -> 10538 input tokens).

## Deliberate gaps (documented in the architecture doc, flagged for owners)

- **Per-search provider fee is not modeled.** Web search costs the provider account $10/1k searches; the gateway bills the grown input-token total but not the per-use fee, and `usage.server_tool_use.web_search_requests` is not yet surfaced. Follow-up needs a pricing-schema decision (the #680 class).
- **Citations are not carried.** `citations_delta` on text blocks is dropped (answer text streams uncited; the result block with URLs IS delivered), and `search_result` input blocks stay rejected. Carrying citations restructures text-block handling end to end and is its own slice.

## Drift gates

Every GA and beta `ToolUnionParam` type literal and every `ContentBlockParam` type must now be a recorded decision; the accepted history-block set is pinned equal to both the wire literal and the native normalizer's Rust constant, so the two engines cannot drift apart.

## Structure

`contracts.py` crossed the 1000-line budget, so replay identity (`provider_replay_authority`, `canonical_request_sha256`) moved to `exp/runtime/gateway/replay_identity.py` (four importers updated, no aliases left behind).

## Gates

- `uv run ruff format` / `ruff check` / `ty check`: clean
- Full `uv run pytest -q` on the committed tree: 2979 passed, 2 skipped
- `cargo test --no-default-features`: 118 passed; `cargo clippy --all-targets`: clean; `cargo fmt --check`: clean
- Crate version bump chores: crate `Cargo.toml`/`pyproject.toml` 0.3.12, `Cargo.lock`, parent pin floor `>=0.3.12,<0.4`, `uv.lock`

### DIFF
--- a/docs/reference/gateway-architecture.md
+++ b/docs/reference/gateway-architecture.md
@@ -237,7 +237,20 @@ bare by the live API, verified 2026-08-30) and `inference_geo` verbatim on Anthr
 disclosure-drops elsewhere, keeps every official SDK tool and top-level field a recorded
 decision behind an SDK-surface drift gate in
 `exp/runtime/anthropic_protocol/manifest.py`, and rejects image and document blocks loudly
-because the surface is text-only. Thinking carriers
+because the surface is text-only. Typed Anthropic tool entries (web search, web fetch, code
+execution, tool search, bash, text editor, memory; every accepted type bare-accepted live,
+verified 2026-08-31) decode onto a verbatim server-tool carrier that re-enters the tools array
+at its exact caller position on Anthropic rungs, and the echoed `server_tool_use` and
+`*_tool_result` assistant history blocks ride an opaque per-block carrier the same way; any
+non-Anthropic rung is a named rejection for both, never a disclosure-drop, because silently
+removing a declared search or execution capability would change what the model can do. On the
+response side the native normalizer folds `server_tool_use` input fragments into one verbatim
+block, delivers result blocks whole, honors the provider's `pause_turn` stop reason, and reads
+the cumulative `message_delta` usage so search-inflated input tokens bill correctly; the
+per-search provider fee itself is deliberately not modeled yet (searched content is billed
+through the grown input-token total). Text citations from search results are not carried yet:
+the answer text streams uncited, and `search_result` input blocks stay rejected for the same
+reason. Thinking carriers
 replay only on the Anthropic wire, so route admission requires every waterfall rung to speak the
 `anthropic_messages` dialect; on the Responses surface over Anthropic routes, thinking text is
 projected onto the reasoning-summary channel (signatures deliberately dropped) so callers receive
--- a/exp/runtime/anthropic_protocol/manifest.py
+++ b/exp/runtime/anthropic_protocol/manifest.py
@@ -137,12 +137,100 @@ rungs, which own their validity rules, and drop with disclosure elsewhere.
 """
 
 MESSAGES_TOOL_FIELDS_REJECTED: frozenset[str] = frozenset()
-"""Custom tool-definition fields consciously rejected, currently none.
+"""Custom tool-definition fields consciously rejected, currently none."""
 
-Server-defined tools (bash, text editor, web search, code execution, tool
-search) are rejected wholesale through the closed ``type`` literal instead
-of field-by-field: the gateway serves no provider-hosted execution.
+
+MESSAGES_SERVER_TOOL_TYPES_ACCEPTED = frozenset(
+    {
+        "bash_20250124",
+        "text_editor_20250124",
+        "text_editor_20250429",
+        "text_editor_20250728",
+        "memory_20250818",
+        "web_search_20250305",
+        "web_search_20260209",
+        "web_search_20260318",
+        "web_fetch_20250910",
+        "web_fetch_20260209",
+        "web_fetch_20260309",
+        "web_fetch_20260318",
+        "code_execution_20250522",
+        "code_execution_20250825",
+        "code_execution_20260120",
+        "code_execution_20260521",
+        "tool_search_tool_bm25_20251119",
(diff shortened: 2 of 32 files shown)


SilenNaihin:
Closing in favor of #690, which was chosen as the shipping implementation for this lane: it additionally carries text citations verbatim end to end (this PR dropped them, acknowledged as a gap), and its conservative accept table admits only tool types whose result-block vocabulary is proven to round-trip.

Ported to #690 as review rather than code: the 22-type live bare-accept probe table (informs the reject rationales), edge-case coverage verification against this PR's independent live capture, and confirmation that the `message_delta` cumulative-usage billing fix there covers all three input legs (both PRs found and fixed the same ~8k-input-tokens-per-search under-billing independently).

Owner-decision gaps flagged here that both implementations share and that remain open after #690 merges:
- The per-search provider fee ($10/1k web searches, charged to the house provider key) is not billed to customers; `usage.server_tool_use.web_search_requests` is parsed past but not surfaced. Needs a pricing-schema follow-up (the #680 class).
- Tool types beyond web_search stay rejected: `bash_20250124` / `text_editor_20250728` / `memory_20250818` emit no new block vocabulary and are candidates for a cheap fast-follow; `web_fetch_*` / `code_execution_*` / `tool_search_*` need their result blocks carried first.

PR #690, opened 2026-08-31
gateway: serve Anthropic server tools (WebSearch) end to end

SilenNaihin:
Serve Anthropic server tools (WebSearch) end to end through the gateway.

## Problem

Claude Code's WebSearch sends a tools entry the strict Messages decoder cannot
represent (`{"type":"web_search_20250305","name":"web_search",...}`, no
`input_schema`), so every session with WebSearch enabled got a 400. Three more
layers sat behind that 400, each verified against the live API with the house
key (2026-08-31):

1. **Stream failure.** The native normalizer skipped `server_tool_use` start
   blocks as unknown, then failed the whole stream at the block's first
   `input_json_delta` ("provider emitted arguments before a tool start").
2. **Turn-2 echo failure.** Multi-turn sessions echo the served blocks back;
   `_REJECTED_BLOCK_HINTS` consciously 400d `server_tool_use` and
   `web_search_tool_result` history blocks, and the strict `_TextBlock`
   rejected the `citations` array every cited answer carries.
3. **Fidelity lies.** `citations_delta` frames were silently dropped, a
   `pause_turn` stop reason would have been rewritten to `end_turn` (the caller
   would never resume the paused turn), and the terminal `message_delta` usage
   (the TRUE post-search input count: 12284 vs 2230 at `message_start` on the
   captured stream, a 5.5x billing undercount) was never read.

## Change

**Request (python decode → canonical contract):**
- `tools` decodes as `_Tool | _ServerTool`; server entries are shallow-validated
  (evolving provider surface) and carried raw on a new Messages-only
  `GatewayRequest.provider_server_tools` carrier that joins replay identity,
  mirroring `provider_thinking_config`.
- Accept/reject decision tables in `manifest.py`
  (`MESSAGES_SERVER_TOOL_TYPES_ACCEPTED/REJECTED`) with a new SDK drift gate
  over `ToolUnionParam`/`BetaToolUnionParam`: all three `web_search_*` versions
  accepted (each verified live to emit the block vocabulary this gateway now
  carries); every other Anthropic-defined type rejected BY NAME because the
  data plane does not carry its result blocks (never a silent half-serve).
- Echoed history blocks (`server_tool_use`, `web_search_tool_result`,
  citation-bearing `text`) ride a new whole-message verbatim carrier
  (`GatewayMessage.provider_anthropic_block`, mirroring `provider_native_item`);
  decode splits the assistant turn at block boundaries and the anthropic
  payload builder merges consecutive assistant messages back, so re-emission
  preserves exact block order. Server tools re-emit after converted custom
  tools (documented ordering deviation).
- Route narrowing rejects server tools on any non-Anthropic rung by name
  (reject-not-drop: silently removing a requested search capability would
  falsify answers). `tool_choice` may name a server tool; a server-tool-only
  toolset no longer no-op-drops `tool_choice`.

**Response (Rust data plane, crate 0.3.11 → 0.3.12):**
- New events: `ServerToolUseStarted/ArgumentsDelta/UseCompleted`,
  `ServerToolResult` (whole verbatim block), `TextBlockStarted` (provider
  text-block boundary so citations attach per block), `CitationDelta`
  (verbatim citation object), and the `PausedTurn` terminal.
- The anthropic normalizer parses all of them, keeps `pause_turn`, and now
  folds the terminal `message_delta` input/cache legs when present (billing
  fix; live plain turns carry the same legs, verified).
- The Messages encoder streams the blocks back in valid Anthropic lifecycle
  order and aggregates them identically for non-streaming callers; server tool
  use never yields the `tool_use` stop reason. Chat/Responses encoders reject
  server-tool events (unreachable by construction) and ignore
  boundary/citation metadata like the thinking family. Guardrail text
  replacement drops server-tool content with the reasoning channel (fetched
  content must not leak through a rewrite). Server tool names join usage
  `tool_names` for per-invocation observability (searches are provider-priced
  at $10/1k; platform-side pricing of search units is a named follow-up).

## Evidence

- Live capture pinned as a parity golden
  (`test_native_anthropic_normalizer_decodes_the_live_web_search_wire`):
  ids neutralized, results trimmed, frame order and shapes are the real wire.
- Live acceptance through a locally served native gateway with the house key:
  turn 1 streamed a real WebSearch 200 end to end
  (`server_tool_use` → `web_search_tool_result` → cited text, stop `end_turn`,
  usage 12282/103), turn 2 echoed the returned blocks and re-served ("3.14.7").
- Turn-2 echo also verified directly against api.anthropic.com with the exact
  captured blocks (200), including the `caller: {"type":"direct"}` extras.
- `web_search_20260209`/`20260318` probed live: identical block vocabulary
  (they 400 on haiku unless `allowed_callers: ["direct"]`; the provider's own
  named, actionable error).

## Gates

- `uv run ruff format` / `ruff check` / `ty check`: clean.
- `uv run pytest -q`: full suite green.
- `cargo test --no-default-features` (118), `cargo clippy -D warnings`,
  `cargo fmt`: clean.

## Known follow-ups (out of scope, named)

- `web_fetch_*` is the obvious next accept: live probes (2026-08-31) show
  `web_fetch_20250910` accepted bare with the same
  `server_tool_use`/`web_fetch_tool_result` lifecycle, so it needs only that
  result-block carriage plus a pinned fixture (bare-accept probes show every
  current-generation type is accepted without beta headers; only the legacy
  2024-10-22 family is beta-gated).
- `bash_20250124`, `text_editor_20250728`, and `memory_20250818` are
  client-executed and emit only ordinary `tool_use` blocks (verified live in
  the #689 probe session), so they are cheap fast-follow accepts once decided
  consciously; they stay rejected by name here.
- On `max_uses` exhaustion the provider replaces the result block's `content`
  list with a single error object. The verbatim carriage handles that shape
  structurally, but no fixture pins it yet; noted as an unpinned case.
- Per-search provider fees are NOT yet customer-billed: Anthropic charges
  $10 per 1,000 searches on the house key, and the gateway's usage contract
  meters tokens only. The `web_search` entry now present in usage
  `tool_names` gives the platform the per-invocation signal to price against;
  the pricing model itself is a platform follow-up for the owner to decide.

### DIFF
--- a/docs/reference/gateway-architecture.md
+++ b/docs/reference/gateway-architecture.md
@@ -237,7 +237,17 @@ bare by the live API, verified 2026-08-30) and `inference_geo` verbatim on Anthr
 disclosure-drops elsewhere, keeps every official SDK tool and top-level field a recorded
 decision behind an SDK-surface drift gate in
 `exp/runtime/anthropic_protocol/manifest.py`, and rejects image and document blocks loudly
-because the surface is text-only. Thinking carriers
+because the surface is text-only. Anthropic server tools are decided per type by the same
+manifest: verified `web_search_*` entries forward verbatim after the converted custom tools,
+their streamed output (`server_tool_use`, `web_search_tool_result`, citation-bearing text
+blocks, and the `pause_turn` stop reason) reaches the caller intact on both response paths, and
+a next-turn echo of those blocks (each carried verbatim as a whole-message block) re-serves
+byte-for-byte; every other Anthropic-defined tool type is rejected by name because the data
+plane does not yet carry its result blocks. Like the thinking carriers, server tools replay
+only on the Anthropic wire, so a route with any other rung rejects them by name instead of
+dropping a requested capability. The terminal `message_delta` usage report supersedes the
+`message_start` input legs when present, because server-tool turns re-read fetched results as
+input and the start-frame count severely undercounts the billed total. Thinking carriers
 replay only on the Anthropic wire, so route admission requires every waterfall rung to speak the
 `anthropic_messages` dialect; on the Responses surface over Anthropic routes, thinking text is
 projected onto the reasoning-summary channel (signatures deliberately dropped) so callers receive
--- a/exp/runtime/anthropic_protocol/manifest.py
+++ b/exp/runtime/anthropic_protocol/manifest.py
@@ -201,13 +201,17 @@ MESSAGES_SERVER_TOOL_TYPES_REJECTED = frozenset(
 """Anthropic-defined tool types consciously rejected by name.
 
 Each is rejected because the gateway cannot yet serve it truthfully, not
-because the provider would refuse it: ``web_fetch_*``, ``code_execution_*``,
+because the provider would refuse it (a live probe on a plain key,
+2026-08-31, accepted the current-generation types bare; beta headers only
+matter for the 2024-10-22 family): ``web_fetch_*``, ``code_execution_*``,
 and ``tool_search_*`` stream result blocks the data plane does not carry
 (silently dropping them would falsify the response); ``code_execution_*``,
 ``browser_toolset_*``, ``computer*``, and ``mcp_toolset`` additionally bind
-provider-hosted execution state; ``computer*``, ``bash_*``, ``text_editor_*``,
-``memory_*``, and ``advisor_*`` are client-executed but unverified here and
-several require beta headers this gateway does not forward. Accepting one
+provider-hosted execution state. ``bash_20250124``,
+``text_editor_20250728``, and ``memory_20250818`` are client-executed
+through ordinary ``tool_use`` blocks (no new block vocabulary in that
+probe), making them the nearest accept candidates once verified end to end;
+the remaining client-executed types are unverified here. Accepting one
 means moving it to :data:`MESSAGES_SERVER_TOOL_TYPES_ACCEPTED` with live
 block-vocabulary evidence, exactly as web_search was.
 """
--- a/exp/runtime/anthropic_protocol/manifest.py
+++ b/exp/runtime/anthropic_protocol/manifest.py
@@ -2,12 +2,12 @@
 
 from __future__ import annotations
 
-from exp.runtime.gateway.contracts import (
+from exp.runtime.gateway.compatibility import (
     CompatibilityDisposition,
     CompatibilityField,
     CompatibilityManifest,
-    GatewayApiSurface,
 )
+from exp.runtime.gateway.contracts import GatewayApiSurface
 
 
 def _field(
@@ -77,8 +77,9 @@ MESSAGES_MANIFEST = CompatibilityManifest(
         #   proxy it truthfully, so it stays rejected until a verified
         #   contract exists.
         # - ``container`` and ``mcp_servers`` bind provider-hosted execution
-        #   state this gateway does not serve (server tools are rejected on
-        #   the same grounds).
+        #   state this gateway does not serve (server tools whose result
+        #   blocks the gateway cannot carry are rejected on the same grounds;
+        #   see MESSAGES_SERVER_TOOL_TYPES_ACCEPTED).
         # - ``service_tier`` selects provider priority capacity priced above
         #   the gateway's committed billing model.
         # - ``fallbacks`` and ``fallback_credit_token`` swap the upstream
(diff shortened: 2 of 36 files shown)


SilenNaihin:
Review from the closed sibling implementation (#689), porting its unique evidence per the lead's decision. Three parts: live probe data for the accept/reject table, edge-case coverage verification, and the usage fix.

## 1. Live bare-accept probe data for the type table (2026-08-31, plain house key, no beta headers, claude-sonnet-4-6)

Every GA `ToolUnionParam` type was probed live with a minimal request. Results:

| Type | Bare result |
|---|---|
| `bash_20250124` | 200 |
| `text_editor_20250728` | 200 |
| `text_editor_20250124`, `text_editor_20250429` | 400, provider's own per-model error ("'claude-sonnet-4-6' does not support tool types: ...") |
| `memory_20250818` | 200 |
| `web_search_20250305` / `20260209` / `20260318` | 200 |
| `web_fetch_20250910` / `20260209` / `20260309` / `20260318` | 200 |
| `code_execution_20250522` / `20250825` / `20260120` / `20260521` | 200 |
| `tool_search_tool_bm25_20251119`, `tool_search_tool_regex_20251119` (with one non-deferred custom tool) | 200 |
| `browser_toolset_20260801`, `computer_toolset_20260801` | 400, provider's own per-model error |

Two consequences for the `MESSAGES_SERVER_TOOL_TYPES_REJECTED` docstring:

- "several require beta headers this gateway does not forward" holds only for the 2024-10-22 family (`bash_20241022`, `text_editor_20241022`, `computer_20241022`; the newer `computer_2025*` variants were not probed). Every current-generation type above is accepted bare on a plain key. Suggest citing the probe rather than the beta-header claim for the current types, so the recorded reason is the durable one (result-block vocabulary not yet carried / unverified), not one that a doc change invalidates.
- `bash_20250124`, `text_editor_20250728`, and `memory_20250818` are client-executed and emit ordinary `tool_use` blocks with no new result-block vocabulary (verified in the same session: the model calls them like custom tools and the caller returns `tool_result`). Under this PR's own bar (truthful end-to-end serving) they appear servable today with zero data-plane work, so they are a cheap fast-follow candidate for the accepted set. Keeping them rejected in this PR is fine; flagging so the table's own principle is applied to them consciously.

## 2. Edge-case coverage check against the sibling's independent live capture

I verified this PR's normalizer and fixtures against the frames captured in the sibling implementation (separate live web_search session, claude-sonnet-4-6). Everything material is covered here: the leading empty `input_json_delta`, the whole `web_search_tool_result` (with `caller`) in its start frame, `citations_delta` arriving BEFORE the first `text_delta` of its span, whitespace-padded frames / `stop_details`, `pause_turn` as a first-class terminal, and the zero-argument seed routed to the server-side event. One shape worth keeping in mind for a regression later (no action needed now): on `max_uses` exhaustion the provider replaces the result block's `content` list with a single error object; the verbatim carry here handles it structurally, but no fixture pins it.

## 3. Usage fix confirmed complete

The `message_delta` superseding covers all three input legs (`input_tokens`, `cache_read_input_tokens`, `cache_creation_input_tokens`) via `count_if_present`, which matches what the sibling fix did. Independent confirmation of the magnitude from the second live capture: `message_start` 2230 -> `message_delta` 10538 input tokens on one single-search turn, and 10038 on a rerun, so every web_search turn was previously under-billing input by roughly 8k tokens.

Shared owner-decision gap carried by both PRs, for the record: the per-search provider fee ($10/1k searches on the house key) is not billed to customers; `usage.server_tool_use.web_search_requests` is parsed past, not surfaced. Needs a pricing-schema follow-up (the #680 class).

SilenNaihin:
Part 1 of the ported #689 evidence is addressed in d4ce4d9d: the rejection-table docstring now cites the 2026-08-31 bare-accept probe instead of the overbroad beta-header claim (which held only for the 2024-10-22 family), and names `bash_20250124`/`text_editor_20250728`/`memory_20250818` as the verified nearest accept candidates. Independent confirmation from a third live session on claude-haiku-4-5: `web_fetch_20250910` bare -> 200 with the same `server_tool_use`/`web_fetch_tool_result` lifecycle, and the same message_delta input supersede magnitude (2288 -> 11108 on one single-search turn). Parts 2 and 3 need no change. The `max_uses`-exhaustion error-content fixture and the `usage.server_tool_use.web_search_requests` pricing gap stay recorded follow-ups.

PR #693, opened 2026-08-31
Demo: Cache-write fields underbilled

FinnBorge:
Hi Experiential team, thank you for the repo/project.  In reading it to learn more about caching, I think I found a gap.  Claude seems to agree.

Anthropic and OpenAI both report a prompt as a total plus a breakdown:
1. Fresh (normal price)
2. Cache-read (discounted)
3. Cache-write (surcharge)

Arithmetic in `budgets.py` inflated totals by summing input + cached_input.  Output + reasoning.
1. 10,000 token prompt served entirely from cache was reported as 20,000 tokens
2. 500 token response that was entirely reasoning was reported as 1,000 tokens

The native `Usage` struct has no field for cache writes, so:
A: 1,000 tokens processed fresh          -> input_tokens=1000, cached_input_tokens=0
B: 1,000 tokens written to the cache     -> input_tokens=1000, cached_input_tokens=0
These are parsed as identical despite B, the cache-write variant, costing more due to cache-write surcharges.

`test_native_anthropic_normalizer_keeps_cache_writes_distinguishable` demonstrates the bug and is intentionally red.

### DIFF
--- a/exp/runtime/gateway/budgets.py
+++ b/exp/runtime/gateway/budgets.py
@@ -826,19 +826,15 @@ def _unknown_token_volume(
 ) -> tuple[int, int]:
     """Sum observed token volume across one limit's unresolved unknown-cost attempts.
 
-    Input volume includes cached input tokens and output volume includes reasoning
-    tokens, so operators can gauge the real traffic behind attempts whose price is
-    still unknown.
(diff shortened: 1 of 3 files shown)


FinnBorge:
Apologies, I should've submitted this as an issue

PR #707, opened 2026-09-01
gateway: the /v1/batches serving lane engine (explicit-request async batches)

SilenNaihin:
## What

The engine half of the /v1/batches product: an asynchronous batch lane where batch models are explicit-request only. One provider serves one whole job, submit is fail-closed (JSONL validation, per-line catalog resolution, single-provider enforcement, per-line reservation before any provider dispatch), and settlement is idempotent per line. Native routes (Rust) arrive in the follow-up PR; this package is fully host-integrable and tested without them.

## Design

- `exp/runtime/gateway/batch/contracts.py`: OpenAI Batch API compatible objects (batch, file, output JSONL), status vocabulary, input parsing with product limits (50k lines / 100MB / 24h window).
- `interfaces.py`: host seams with no implementations, mirroring the gateway injection idiom: `BatchStore`, `BatchFileStore`, `BatchCatalog` (resolves ONLY explicit batch models; a synchronous name returns None and is refused per line), `BatchLedger` (reserve at submit, settle at retrieval, release on expiry/cancel: the existing money mechanics per line), `BatchSecretResolver`.
- `providers.py`: OpenAI (files + batches + output/error files), Anthropic (message batches; the results URL is exact-host validated before any credentialed fetch), OpenRouter (beta batches; inline results; uniform-model required; cancel refused as unsupported). No dialect translation: a line's body is already shaped for the surface it names.
- `engine.py`: submit/validate/reserve (rollback on any rejection), poller (submit deferred to first poll so a caller-visible job always exists before provider state), idempotent per-line settlement into output/error files, expiry and cancellation releases.
- `plane.py`: `BatchControlPlane` with bridge-callable string-JSON methods (`batch_create/retrieve/list/cancel`, `file_create/retrieve/content`) beside the synchronous control plane, authenticating through the host's `GatewayControlStore`.

## Pricing and money

Batch list price passes the provider discount through to the caller: hosts author batch rates on their batch catalog rows; the engine reserves and settles against the host ledger with the existing reserve/settle semantics, one ledger entry per line.

## Testing

- Wire-level provider matrix over `httpx.MockTransport` (real bytes, no network): submit/poll/results/cancel for all three providers across success, per-line error, job failure, spoofed-URL, and unsupported-cancel outcomes.
- Engine lifecycle over in-memory seams: validation matrix, reservation rollback, submit-poll-settle idempotence, expiry, pre-dispatch cancel, owner scoping.
- Control-plane envelope tests: auth, error-status mapping, base64 file round trip.
- Live, credential-gated (`EXP_LIVE_BATCH=1` + provider key): one tiny real two-line batch per provider, submit -> poll -> retrieve -> per-line assertions, in `batch/tests/live_test.py`. Not exercised in this environment: no provider keys are present locally (house keys live in the platform's deployment secrets); the platform-side integration will run them against real endpoints.

## Follow-up (PR 2)

Native routes `/v1/batches*` + `/v1/files*` in the Rust server as thin relays onto these plane methods, sync-route refusal of batch-only model ids with a pointer to /v1/batches, and the wiring into `NativeGatewayComponents`.

Gates: ruff check/format clean, ty clean, full pytest 3044 passed / 5 skipped (the two w16 evidence tests fail only on checkouts with untracked files; green after commit).


SilenNaihin:
Both P1s fixed in e8300d3f, each with a pinned regression test:

1. **Provider submission is not recoverable**: dispatch now records intent (`dispatch_started`) durably before the provider call; a poller that finds intent without a `provider_batch_id` fails the job closed with an operator-actionable message and releases every reservation, never resubmitting paid work against an unknown provider-side batch. Test: `test_interrupted_dispatch_fails_closed_without_resubmitting` (asserts zero provider submits and full release).

2. **Running jobs lose credential binding**: `credential_reference` freezes on the job at submit; `_api_key` resolves only that frozen reference (the secret value stays late-bound through the resolver). Catalog edits mid-job can no longer repoint an open job at a different connection. Test: `test_open_jobs_use_the_submit_time_credential_reference` (repoints the catalog mid-job, asserts the original key is used).

Gates re-run: ruff/format/ty clean, 43 package tests green.

SilenNaihin:
Round-2 P1s fixed in a58f439b, each with a pinned test:

1. **Submission failures become ambiguous**: a raised `BatchSubmitError` from submit means a provider response was received, so nothing was accepted: the job terminalizes immediately with the provider's reason (`test_definitive_submit_rejection_fails_immediately_with_the_reason`). Ambiguous transport losses (any other exception) still take the fail-closed interrupted path on the next poll, which is the deliberate no-double-spend posture.

2. **Credential binding is non-atomic**: `_validate_lines` now returns the deployment binding captured from the same resolution it validated with; the second lookup is gone, so validation and binding cannot be split by a concurrent catalog edit (`test_validation_and_binding_share_one_catalog_resolution`: a catalog that repoints after the first lookup, asserting one lookup and the original reference frozen).

Gates: ruff/format/ty clean, 45 package tests green.

SilenNaihin:
Addressing the Superagent scan's three areas explicitly (6c1c80a1):

- **Account isolation**: every public read and write is owner-scoped through the store contracts (`load_job`/`load_metadata`/`load_content`/`list_jobs` all take the caller's organization), the plane authenticates each call through the host's `GatewayControlStore`, and owner scoping is test-pinned (`test_list_jobs_is_owner_scoped`, the 404 plane tests). Output files are written under the job's own organization.
- **Concurrency**: the engine now STATES its single-writer contract: exactly one poller per job store; hosts with multiple workers lease the poller role. The dispatch-intent guard fails an interrupted dispatch closed; settlement re-loads persisted state before settling; and re-settles are absorbed by the host ledger's contractual idempotency (`BatchLedger` docstrings).
- **Partial settlement**: a crash mid-settle re-enters through the same idempotent path: unsettled jobs re-settle from persisted state, missing lines release with the terminal reason, and the job marks settled only after every line was settled or released, so no upstream work can end both unbilled and unreleased.

Scan verdict is advisory (no branch protection requires it; it skipped on prior merged PRs); the concerns are answered in code and contract above.

SilenNaihin:
All six round-3 findings fixed structurally in cc77918b (each with a pinned test):

- **Atomic dispatch claim** (`BatchStore.begin_dispatch`, one compare-and-set): exactly one actor ever owns a job's provider submission. This closes BOTH the concurrent double-submit (brin) and the cancellation-versus-dispatch race (Greptile): a cancel that wins the claim proves no submission ever ran and releases safely; one that loses marks CANCELLING without touching money, and a post-submit CANCELLING is detected and forwarded to the provider. Tests: `test_cancel_during_inflight_dispatch_keeps_reservations`, `test_interrupted_dispatch_fails_closed_without_resubmitting`.
- **Connection uniformity per line** (both credential threads): every line must share the job's provider connection, not merely its provider; mismatches split out per line as `connection_mismatch`. Test: `test_same_provider_different_connections_split_per_line`.
- **Ambiguous 2xx responses** are `AmbiguousProviderResponse`, never a rejection: an accepted-but-unparseable submit takes the fail-closed interrupted path instead of double-releasing. Test: `test_ambiguous_submit_response_takes_the_fail_closed_path`.
- **Partial terminal settlement**: cancelled/expired/failed dispatched jobs fetch and settle whatever partial results the provider reports, releasing only lines that produced nothing. Test: `test_terminal_jobs_settle_partial_provider_results`.

Gates: ruff/format/ty clean, 49 package tests green.

SilenNaihin:
Round-4 P1s fixed (with tests):

1. **Cancellation intent gets overwritten**: CANCELLING never downgrades to a non-terminal snapshot status, and the in-flight-cancel branch forwards to the provider only when `supports_cancel` (OpenRouter jobs keep the intent and run to their provider-terminal state: the strongest cancellation that provider offers, then settle normally). Test: `test_inflight_cancel_without_provider_support_runs_to_terminal`.
2. **Terminal settlement cannot resume**: the `open_jobs` contract now explicitly includes terminal jobs whose settlement has not completed (the engine already handled them; the protocol wording would have let a host exclude them). Test: `test_interrupted_terminal_settlement_resumes_from_open_jobs`.

Gates: ruff/format/ty clean, 51 package tests green.

SilenNaihin:
Round-5 P1 fixed in 88505d23: a completed job's failed results fetch now propagates and settlement retries on a later poll (settlement is resumable per the open_jobs contract), never releasing lines whose provider work already ran. Release-all remains only for a definitive no-results response on failed/expired/cancelled jobs; ambiguous responses always retry. Test: `test_completed_job_settlement_retries_on_fetch_failure` (first fetch 500s: job stays unsettled with zero releases; second fetch settles the line with zero releases). Gates: ruff/format/ty clean, 52 package tests green.

SilenNaihin:
Round-6 P1 fixed in 8a270ce4: the claim-rejected branch re-loads persisted state before concluding anything and returns when the claim holder already resolved the job (cancel-settled, id persisted, terminal, or CANCELLING); only a persisted claim without a provider id fail-closes, finalizing the reloaded record rather than the stale snapshot. Test: `test_poller_with_stale_snapshot_never_overwrites_a_cancelled_job` (cancel wins the claim, a poller holding the pre-cancel snapshot advances: status stays CANCELLED, settled stays true, zero extra releases, zero submits). Gates: ruff/format/ty clean, 53 package tests green.

SilenNaihin:
Round-7 P1 fixed in 2fcc3a29 (+import restore 1bf01ca2): CANCELLING persists BEFORE the provider cancel call, so a transport or provider failure cannot lose the caller's intent; the poller re-requests provider cancellation on every pass (suppressing transient failures) until the provider reports a terminal state. Test: `test_failed_provider_cancel_keeps_the_intent_and_retries` (first provider cancel 500s: status stays CANCELLING; the next poll re-requests, the provider terminalizes CANCELLED, settlement releases once). Gates: ruff/format/ty clean, 54 package tests green.

PR #711, opened 2026-09-01
gateway: native /v1/batches and /v1/files routes (batch lane PR 2)

SilenNaihin:
## What

The Rust half of the /v1/batches product (PR 2 of the series; the engine landed in #707): native routes for `POST/GET /v1/batches`, `GET /v1/batches/{id}`, `POST /v1/batches/{id}/cancel`, `POST /v1/files` (multipart), `GET /v1/files/{id}` and `/content`, each a thin relay shaping one keyed JSON argument onto the bridge's batch relay methods, whose `{status, body}` envelopes map directly onto HTTP.

## Wiring

- `NativeGatewayComponents` gains an optional `batches` member (a `BatchControlPlane`); the local launch sets it to `None` and a composition without it answers every batch route with a uniform not-enabled error.
- `NativeBatchRelayMixin` (new module, keeps `native_bridge.py` under the line cap) holds the seven relay methods; `NativeControlPlane` inherits it.
- Explicit-request enforcement, sync side: a batch-only model id on `/v1/chat/completions` (or any sync surface) is refused 404 with "only available through the Batch API. Submit it explicitly via /v1/batches" (mirrors the upstream did-you-mean style).
- Cargo: axum `multipart` feature + `base64`; upload transport cap 110MB; downloads stream exact stored bytes as `application/jsonl`.

## Testing

Wire-level certification (`exp/runtime/gateway/tests/native_batches_test.py`): a real native server on a loopback socket over real gateway components (real SQLite authority, real virtual key), batch lane on in-memory seams with a scripted provider: full lifecycle (multipart upload -> submit -> poll -> settle -> output download with exact bytes), owner-scoping refusals, wrong-key 404, the sync-route batches pointer, and the not-enabled composition. Plus the engine package suite (58 tests) and the full repo suite green (3059+ passed; cargo fmt/clippy clean; wheel builds via maturin, 0.3.15).

## Live verification (series note)

The gated live suite ran against the real Anthropic Message Batches API with the locally available key: one two-line haiku batch, submit -> poll -> complete -> per-line results, passed in 5m34s. OpenAI and OpenRouter keys are not present in this environment; their gated tests are ready to run where those keys exist.

### DIFF
--- a/exp/runtime/gateway/batch/engine.py
+++ b/exp/runtime/gateway/batch/engine.py
@@ -98,6 +98,10 @@ class BatchEngine:
         self._clients = clients if clients is not None else PROVIDER_CLIENTS
         self._poll_interval = max(1.0, poll_interval_seconds)
 
+    def is_batch_model(self, *, model: str) -> bool:
+        """Return whether the catalog resolves one explicit batch model."""
+        return self._catalog.batch_deployment(model=model) is not None
+
     def upload_file(
         self, *, organization_id: str, filename: str, purpose: str, content: bytes
     ) -> BatchFile:
--- a/exp/runtime/gateway/batch/engine.py
+++ b/exp/runtime/gateway/batch/engine.py
@@ -372,13 +372,19 @@ class BatchEngine:
             return job
         client = self._clients[job.provider]
         if job.provider_batch_id is not None:
+            if not client.supports_cancel:
+                # An honest refusal beats a silent no-op: the caller learns
+                # the provider limitation and the job state stays untouched.
+                raise BatchSubmitError(
+                    f"{job.provider} batches cannot be cancelled; the job runs to completion",
+                    code="cancel_unsupported",
+                )
             # The intent persists before the provider call, so a failed call
             # cannot lose it: the poller re-requests cancellation until the
             # provider confirms a terminal state.
             job = job.model_copy(update={"status": BatchStatus.CANCELLING})
             self._store.save_job(job=job)
-            if client.supports_cancel:
-                await client.cancel(job=job, api_key=self._api_key(job))
+            await client.cancel(job=job, api_key=self._api_key(job))
         elif self._store.begin_dispatch(batch_id=job.batch_id):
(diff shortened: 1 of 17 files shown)


SilenNaihin:
All three round-1 P1s fixed in 6a1b0e6c, each with wire-level coverage:

1. **Authority failures expose batch models**: the batch pointer now fires only for authenticated identities (the mapped authority error's status is inspected; 401s keep their generic envelope), so an invalid key cannot enumerate batch models. Test: `test_invalid_keys_never_learn_that_a_model_is_batch_only` (asserts 401 without the pointer text).
2. **Multipart errors silently discarded**: multipart reads now answer honestly per failure (413 for the transport cap via `MultipartError::status`, 400 for malformed/truncated framing) instead of collapsing into the missing-file message.
3. **Uploads allocate before authentication**: `/v1/files` and `/v1/batches` submissions authenticate over the bridge BEFORE any body is buffered, exactly like the synchronous routes' pre-auth; the same test asserts an invalid-key upload refuses 401 (before body processing).

Gates: cargo fmt/clippy clean, wheel rebuilt (0.3.16), ruff/format/ty clean, 169 gateway tests green.

SilenNaihin:
Round-2 P1s (both the aggregate-admission class) fixed in the latest push: every batch handler acquires one owned permit from the same semaphore that bounds synchronous requests and holds it for the handler's duration, so batch bodies and uploads share the plane's aggregate concurrency/memory budget and cannot bypass admission. At capacity: honest retryable 429; draining: the draining error. Verified: cargo fmt/clippy -D warnings clean, wheel rebuilt, all wire-level batch tests green (the permit path is exercised by every existing e2e call).

SilenNaihin:
Round-3 P1 fixed in 5738db8c: all seven batch handlers (including downloads and the read-only relays) now hold one aggregate admission permit for their duration — no batch route can buffer content outside the plane's shared concurrency/memory budget. cargo fmt/clippy -D warnings clean, wheel rebuilt, wire tests green.

PR #728, opened 2026-09-02
feat:Add service_tier passthrough for openai_compatible chat completions

jaedmunt:
Hey!

This PR adds service_tier passthrough for openai-compatible providers. 

Right now the gateway rejects service_tier outright, so there's no way to reach a provider's cheaper tiers (like Doubleword's flex, ~25% off, 1 min TTFT and then high-throughput). This PR lets the field through and forwards it verbatim to the provider on both the Chat Completions and Responses surfaces.

Flex is the focus for us but rather than hard-code it to flex only, we accept the full set of OpenAI service_tier values (auto, default, flex, scale, priority, fast). That keeps it matching the OpenAI SDK so normal SDK requests don't get falsely rejected, and it works for any openai-compatible provider, not just ours. The gateway stays a passthrough and forwards whatever the caller sends, and the provider decides what it honors.

The changes are Python-only as the outbound request body is built in Python and Rust forwards it as-is, so no native changes were needed. It's a small thread-through to accept the field at decode, carry it on the canonical request, and emit it in the outbound payload.

Scope note. this is just the service_tier param. It doesn't touch Responses background mode or add a batch endpoint, those are separate pieces. 

At least for batches, we're keen to get this intergrated, but this is held out for a separate PR, and not included in this one :) 

Jaedon Munton
Member of Technical Staff @ Doubleword


kfallah:
Pushed a follow-on commit (65f5f331) onto this branch (maintainer edit) that adds the route-admission and disclosure layer the review asked for: dialects without a service_tier wire field (Anthropic, Gemini, Bedrock) now decline the rung so admission prefers a tier-preserving route, a route with no preserving rung serves with the tier dropped and disclosed via ignored_parameters, and both manifests reclassify service_tier as CONDITIONALLY_SUPPORTED. Your two commits are untouched underneath — thanks for the contribution!

SilenNaihin:
Full transparency on a process error I made on this branch today, plus review findings and an offered patch.

**What happened**: while rebuilding this PR against current `main` I force-pushed my stack to `doubleword-flex` under maintainer-edit rights. My first push was rejected by the lease — @kfallah had pushed his own reshape in the meantime — and I wrongly treated that as staleness and pushed over his commits. That was my mistake: a rejected push is a stop signal, and I should have read the unfamiliar commits and stopped. **His stack is restored unchanged** (`3b0829d5` is the branch tip again; nothing was lost, my overwrite lasted ~30 minutes) and I will not push to this branch again.

**My parallel reshape is published separately** at [`service-tier-byok-gate`](https://github.com/experientiallabs/experiential/tree/service-tier-byok-gate) (`db92f2a6`, builds on jaedmunt's two commits with authorship preserved; full suite 3261 passed, ruff/ty clean, live-verified on a BYOK OpenAI rung). Offered as a patch to pull from, not a competing takeover — @kfallah's mixed-waterfall coerce-drop structure (`coerce_route_rejections`) and the admission tests here are good and worth keeping.

**Review findings on the current branch state, from the owner's spec for this feature**:

1. **BYOK-only billing gate is required and missing.** The current dialect-only gate (`SERVICE_TIER_DIALECTS`) forwards `service_tier` on host-funded rungs, where the tier changes what the provider charges (flex discounted, priority premium) while the gateway bills catalog rates — a silent billing mismatch. The funding fact is already frozen per deployment (`billing_source`); my branch binds it into `GatewayWireProfile.billing_customer_managed` at wire resolution and gates emission on it (`execution_resolution.py`, `streaming_requests.py`).
2. **Digest safety**: a plain `GatewayRequest.service_tier` field changes `canonical_request_sha256` for every existing request (serialization uses `exclude_none=False`), breaking replay identity for tier-free traffic. It needs `Field(..., exclude=True)` plus an only-when-present fold into the replay envelope — same tier + same body must replay, different tier must conflict (my branch: `contracts.py`, `replay_identity.py`, pinned by `test_service_tier_is_serialization_inert_but_binds_replay_identity`).
3. **`"fast"` is not an OpenAI service tier** (it's Anthropic's speed selector); keeping it in the Literal accepts a value every OpenAI-family upstream rejects. My branch trims the enum to `auto/default/flex/scale/priority`.
4. One behavioral difference worth a deliberate choice: raising `ProviderCapabilityError` for the tier per-rung narrows host-funded rungs out of tier-carrying routes entirely, so a waterfall whose BYOK rung is down 400s instead of serving untiered. The spec asks for strip-and-serve with an `ignored_parameters` disclosure on ineligible rungs (my branch does the structural drop + route-level disclosure).

Happy to reconcile whichever direction @kfallah prefers — pulling the gate/digest/enum commits from `service-tier-byok-gate` onto this branch keeps his structure and closes the spec gaps.

jaedmunt:
Thanks for reviewing (no worries about the push :) )

kfallah:
@SilenNaihin please free to bring ur changes in here and land. thanks again @jaedmunt !

SilenNaihin:
Takeover announcement, per @kfallah's request (relayed by Silen: "can you please takeover? seems like there are some necessary changes").

Plan before any push, so there are no more surprise rewrites on this branch:

- Base stays **his current tip `3b0829d5`** — his and @jaedmunt's commits and authorship are preserved; my changes land as commits on top.
- **BYOK funding gate**: bind `billing_customer_managed` into the wire profile from the frozen deployment's `billing_source`; house-funded rungs never emit the tier. His per-rung `ProviderCapabilityError` raise (which narrows house-funded rungs out of tier-carrying routes) becomes a structural non-emission, and routes where NO rung may forward the tier flow into his `coerce_capability` drop path so his `SERVICE_TIER_DROP_DISCLOSURE` fires — strip-and-serve, not availability narrowing.
- **Replay identity**: the plain contract field becomes a serialization-excluded carrier folded into `canonical_request_sha256` (historical digests unchanged when absent; same body at a different tier is a different operation), with the digest-stability test.
- **Enum**: `"fast"` removed from both wire models (it is Anthropic's speed selector, not an OpenAI tier). His manifest `CONDITIONALLY_SUPPORTED` moves stay.
- **Tests**: his suite kept green (expectations adapted where raise becomes drop, called out in the commit message), plus the funding-gate/emission/replay tests and a live BYOK wire-presence verification.

One clean push after local gates (full suite + ruff + ty), then a delta-table comment here. `service-tier-byok-gate` stays up for reference until this merges.

SilenNaihin:
Takeover executed as announced: four commits fast-forwarded onto @kfallah's tip `3b0829d5` (no history rewritten; his and @jaedmunt's commits stand as authored). Delta versus his tip:

| Change | What | Why |
|---|---|---|
| `da8f0b3e` enum trim | `"fast"` removed from both wire Literals (now `auto/default/flex/scale/priority`), decode + rejection tests on both lanes | `fast` is Anthropic's speed selector, not an OpenAI tier; every OpenAI-family upstream rejects it |
| `2737e3ef` replay identity | `GatewayRequest.service_tier` becomes serialization-excluded and folds into `canonical_request_sha256` only when present, digest-stability test added | a plain field changes every historical request digest (`exclude_none=False` serialization); and the same body at a different tier is a different provider price/schedule, so it must be a different operation |
| `f7e7f823` BYOK gate | `GatewayWireProfile.billing_customer_managed` bound from the frozen deployment's `billing_source`; the two tier-preserving builders emit only on BYOK rungs; house-funded OpenAI rungs serve untiered and STAY in the route (structural non-emission, no decline); route shaping strips the tier with the same `service_tier` disclosure when no BYOK tier-preserving rung exists | the owner-required billing guard: a tier changes what the provider charges (flex discounted, priority premium) while the gateway bills catalog rates, so it forwards exactly where the caller pays the provider directly |
| `188aa8d1` module split | the two OpenAI builders move to `openai_payloads.py` (mirrors the existing `messages_payloads` split); no behavior change | `streaming_requests` crossed the 1000-line gate |

Kept from the existing stack, deliberately: the dialect decline in `dialect_stream_payload` (still guards mixed routes where an eligible rung keeps the tier alive), the `coerce_route_rejections` mixed-rejection drop path and `SERVICE_TIER_DROP_DISCLOSURE`, and the manifest `CONDITIONALLY_SUPPORTED` classification. One test adaptation, called out in the commit message: the mixed-waterfall coercion test's shim rung is now BYOK so the tier survives route shaping and the coercion path remains the exercised mechanism (a house-funded shim is covered by the new route-shaping test). Two duplicate manifest entries from the merge were deduped.

Verified: full suite 3269 passed / 6 skipped, ruff + ty clean, and live through a locally served gateway on a BYOK OpenAI rung — `service_tier: "default"` serves, and the differential proof that the tier reaches the wire: `flex` on a non-flex model draws the upstream rejection while the same request without the tier serves.

Not in scope, as noted in the plan: Anthropic's own `service_tier` (stays a recorded Messages-surface rejection) and response-side tier echo into usage events (follow-up; the BYOK-only gate makes the billing-audit need non-urgent).

jaedmunt:
Thank you!

PR #740, opened 2026-09-03
gateway: pin the upstream-abort hardening invariant (truncation matrix + h2 abort tests)

SilenNaihin:
## Context

Prod gateway-worker pods are dying time-correlated with the deepseek-v4-flash house lane's fronting proxy logging "aborting with incomplete response" mid-stream (4 deaths across 20 aborts in 14h, all on deployment mp-1f2496a4, terminal_state=unknown_after_crash). Team-lead ordered a repro harness at controlled byte offsets through the native data plane, a fix if a killing offset class exists, and a committed hardening matrix regardless.

## What the hunt found

- **h1 truncation is exonerated.** An `.agents/` sweep (not committed) drove ~8,000 scenarios against a real serving subprocess on 0.7.25: every byte offset of a deepseek-shaped tool-call SSE response (872B, then 2,606B with streamed tool arguments and raw UTF-8) x FIN and RST closes x streamed/non-streamed x slow-reading clients x 16-way concurrent abort storms x delayed mid-stream aborts. Zero process deaths, zero health-probe failures; every abort classified (502 `all_routes_failed` or served-partial 200).
- **The production wire is HTTP/2.** Live ALPN probe of the house lane's base URL negotiates `h2` over TLSv1.3, so prod aborts arrive as `RST_STREAM`/`GOAWAY` frames — a different reqwest/hyper path than h1 truncation, and one h2 connection multiplexes many streams (matches the whole-pod blast radius).
- **The relay layer handles both h2 abort shapes correctly** (proven by the new Rust tests below): delivered deltas surface first, then a classified, failover-eligible `Transport` failure. No panic.
- The remaining crash hypotheses (hyper h2-internal path above the relay, OOM, liveness-probe kill) need the pod's previous-termination evidence, which team-lead is pulling. If that surfaces a real vector, the fix rides a follow-up; this PR pins the invariant the incident violates so it can never regress silently.

## What this ships

1. **`exp/runtime/gateway/tests/native_truncation_test.py`** — end-to-end truncation matrix. A raw-socket chunked-SSE upstream truncates at one representative offset per class (nothing / mid-status-line / mid-headers / headers-only / mid-chunk-length-header / mid-first-frame / mid-raw-UTF-8-codepoint / between-frames / mid-tool-call-arguments / before-final-chunk / complete) x FIN and RST x three surfaces (chat-stream, chat-nonstream, responses-stream), 66 scenarios against a real `exp_gateway_native.serve` subprocess. Every request must return 200 or 502, `process.poll()` must stay alive after every single request, and `/health/live` must be green at the end.
2. **`relay.rs` h2 abort regression tests** — an h2c frame-level server (new dev-dependency `h2 = "0.4"`) sends two SSE DATA frames then either `RST_STREAM(INTERNAL_ERROR)` or a whole-connection drop while the client is mid-body. Both must yield the delivered `TextDelta`s followed by a failover-eligible `FailureClass::Transport`, never a panic.
3. Crate 0.3.22 -> 0.3.23 (test-only change: dev-dependency + `#[cfg(test)]` modules; no data-plane behavior change) with the usual pin/lock chores.

## Validation

- `uv run pytest -q` on the committed tree: 3376 passed, 6 skipped
- `cargo test --no-default-features` (PYO3_PYTHON pinned to the worktree venv): 145 passed, incl. the 2 new h2 tests
- `cargo clippy --all-targets --no-default-features`: clean; `cargo fmt --check`: clean

Target: next release train.

### DIFF
--- a/exp/runtime/gateway/native/Cargo.lock
+++ b/exp/runtime/gateway/native/Cargo.lock
@@ -212,12 +212,13 @@ dependencies = [
 
 [[package]]
 name = "exp-gateway-native"
-version = "0.3.22"
+version = "0.3.23"
 dependencies = [
  "axum",
  "base64",
  "bytes",
  "futures-util",
+ "h2",
  "http",
  "http-body-util",
  "pyo3",
--- a/exp/runtime/gateway/native/Cargo.toml
+++ b/exp/runtime/gateway/native/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "exp-gateway-native"
-version = "0.3.22"
+version = "0.3.23"
 edition = "2021"
 description = "Rust data plane for the Experiential local gateway (PoC), embedded as a PyO3 extension."
 license = "Apache-2.0"
@@ -31,6 +31,12 @@ serde = { version = "1", features = ["derive"] }
 serde_json = { version = "1", features = ["preserve_order"] }
 sha2 = "0.10"
 thiserror = "2"
+
+[dev-dependencies]
+# Frame-level HTTP/2 server control for the reset/goaway regression tests:
+# the production wire to fronting proxies negotiates h2, so mid-stream aborts
+# arrive as RST_STREAM/GOAWAY, a different reqwest path than h1 truncation.
+h2 = "0.4"
 
 [profile.release]
 lto = "thin"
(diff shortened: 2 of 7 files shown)


SilenNaihin:
Addressed: the matrix no longer tolerates any client-side httpx transport/protocol exception — a worker that severs or corrupts its own response now fails the test directly, matching the stated 200/502 invariant. All 66 scenarios still pass with the strict form.

PR #741, opened 2026-09-03
gateway: make catalog identity additive-stable (exclude-default serialization, schema v3)

SilenNaihin:
## DO NOT MERGE until the platform's skew-tolerant published-snapshot reader is confirmed live in production

This change deliberately moves every catalog digest one final time (schema v3). It must roll only into a fleet whose platform reader tolerates a pinned digest it cannot re-derive; team-lead confirms when that hotfix is deployed.

## Incident this ends

The published catalog snapshot failed digest verification on every worker read from 09-02 07:22Z. Root cause (catalog-heal's byte-level repro + this audit): `NormalizedGatewayCatalog`'s identity serialization changed in three consecutive releases without a `SNAPSHOT_SCHEMA_VERSION` bump, so mixed-pin fleets failed the same-version strict digest check on every read:

| Release | Identity change | Bump? |
|---|---|---|
| 0.7.22 | `GatewayDeploymentCapabilities` + `supports_image_input`, `supports_image_url_input` | no (stayed 1) |
| 0.7.23 | + `supports_video_input`, `supports_video_url_input`, `supports_pdf_input`, `supports_pdf_url_input` | no |
| 0.7.24 | none | — |
| 0.7.25 | + `supports_audio_input` (audited: this is the whole identity delta, the third distinct digest) | no |

Each new `bool = False` serialized (`exclude_none=False`) into all ~3k deployments — thousands of leaf diffs, new digest, unchanged version. The existing change-detector (`test_normalized_schema_change_requires_a_schema_version_bump`) fingerprints only the TOP-LEVEL field names of the three normalized models, so nested growth never tripped it.

## The fix: additive fields can no longer change identity

`identity_sha256()` and `_capability_declaration_sha256()` now serialize with `exclude_defaults=True` (recursive, `default_factory`-aware). A field holding its declared default contributes zero identity bytes, so:

- adding a defaulted field anywhere in the catalog graph reproduces every existing digest and every singleton `exact_model_id` unchanged — no bump needed, ever, for additive growth;
- an explicitly spelled default hashes identically to leaving the field unset (contract equivalence);
- `schema_version` itself is default-excluded from identity, so future bumps also stop moving digests; roll-skew detection is unaffected because stored snapshot files keep the FULL dump with `schema_version` persisted (`is_foreign_snapshot` reads the stored value).

Consequence: snapshot files are no longer byte-addressed by name (name = identity digest, bytes = full dump). Nothing in the codebase relied on that byte equality except one test assertion, now replaced with the pinned-reader round-trip; `read_pinned_normalized_snapshot`, `budgets`, `sqlite.platform`, `routing`, and `lifecycle` all verify identity from the parsed model already.

`SNAPSHOT_SCHEMA_VERSION` 2 -> 3: this PR is itself an identity change and follows its own rule (2 is unreleased, but previews may have run main).

## The detector that makes silent violation impossible

`test_identity_digest_is_pinned_until_a_deliberate_schema_version_bump`: a frozen fixture catalog (nested non-default capability flags, integer prices + long-context tier, authored capabilities, both billing sources, certified `maximize_cache` pool, bare all-default record) must produce exactly the pinned `(SNAPSHOT_SCHEMA_VERSION, identity_sha256)` pair. Two lawful outs, stated in the failure text: default the new field (identity ignores it — the test passes untouched), or bump and repin together. A companion test pins the additive-stability property itself (every default-constructed nested declaration dumps to `{}` under identity serialization, and an explicitly spelled default catalog hashes equal). The top-level shape fingerprint test remains for structural changes.

## Validation

- Full `uv run pytest -q` on the committed tree: 3377 passed, 6 skipped
- No `native/` changes; no crate bump needed

## Finding 6 (routed from platform #1156's review): the authored literal was the same timebomb

`ModelCatalog.schema_version` was `Literal[2]` — and the authored document is what every cross-version hydration parses FIRST. A future literal bump raises `literal_error` on older pods, which `load_forward_compatible` cannot drop (it only removes unknown extras), so the platform tolerance never gets a parsed document to serve: warm-fatal on old pods plus the heal-pass mass-deactivation hazard return, relocated one document over.

Fixed here by de-literaling: `schema_version` is now a sane-ranged int (`ge=2, le=10_000`, mirroring `is_foreign_snapshot`'s posture), so additive authored revisions are roll-safe by construction; the field docstring pins the rule that a revision reinterpreting EXISTING fields must use a new field name or a fleet-first tolerance release. Version 1 stays rejected at validation (migration-path only) and the bool/float lookalike validator stays. A dedicated change-detector pins the acceptance window and the authored top-level field fingerprint. No identity movement — the pinned normalized fixture digest is untouched.

### DIFF
--- a/exp/common/models/catalog.py
+++ b/exp/common/models/catalog.py
@@ -719,10 +719,32 @@ class ModelRoles(ContractModel):
         return self
 
 
+SANE_MAX_MODEL_CATALOG_SCHEMA_VERSION = 10_000
+"""Upper bound on an authored catalog version this parser accepts as real.
+
+Mirrors the normalized snapshot's sane-range posture: no product will ever ship
+this many authored-catalog schema revisions, so a value beyond it is corruption
+and fails closed rather than being read as a future contract.
+"""
+
+
 class ModelCatalog(ContractModel):
     """The local model aliases, connection metadata, and project role assignments."""
 
-    schema_version: Literal[2] = 2
+    schema_version: int = Field(default=2, ge=2, le=SANE_MAX_MODEL_CATALOG_SCHEMA_VERSION)
+    """Authored catalog contract revision. Deliberately NOT a ``Literal``.
+
+    Every cross-version hydration parses the authored document first, and a
+    changed ``Literal`` value on a known field raises ``literal_error``, which
+    the forward-compatible read path cannot drop — so a literal here makes any
+    future authored revision warm-fatal on every older pod (the same outage
+    class as the 09-02 catalog incident). A newer stamp within the sane range
+    parses under this build's semantics instead. That makes additive revisions
+    safe by construction; a revision that REINTERPRETS existing fields must not
+    reuse this channel — it needs a new field name or a fleet-first tolerance
+    release. Version 1 stays rejected here: it is only readable through
+    ``_migrate_legacy_model_catalog`` on the TOML load path.
+    """
     connections: dict[str, ConnectionConfig]
     models: dict[str, ModelRecord]
     gateway_pools: dict[str, GatewayPoolRecord] = Field(default_factory=dict)
--- a/exp/common/models/catalog_test.py
+++ b/exp/common/models/catalog_test.py
@@ -25,7 +25,10 @@ from exp.common.models import (
     load_model_catalog,
     write_model_catalog,
 )
-from exp.common.models.catalog import infer_azure_api_surface
+from exp.common.models.catalog import (
+    SANE_MAX_MODEL_CATALOG_SCHEMA_VERSION,
+    infer_azure_api_surface,
+)
 
 
 def _catalog() -> ModelCatalog:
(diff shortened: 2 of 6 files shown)


SilenNaihin:
Heads-up: #744 (`reasoning_tokens_additive` on `GatewayDeploymentCapabilities`, the per-deployment declaration for xAI / Azure Foundry Grok lanes that report reasoning outside `completion_tokens`; platform#1157 step 1-3) is stacked on this branch and held behind it.

Why here: on `main`'s identity serializer a new defaulted capability field moves every deployment digest (the same mechanism as the 0.7.22 -> 0.7.25 incident), so it cannot land there without a schema bump, and the bump is this PR's. On this branch the field is digest-neutral by construction: `GatewayDeploymentCapabilities().model_dump(exclude_defaults=True) == {}` with the field present, and `test_identity_digest_is_pinned_until_a_deliberate_schema_version_bump` / `test_added_defaulted_fields_and_explicit_defaults_do_not_perturb_identity` pass untouched on #744. That is exactly the additive-growth contract this PR states.

Plan: #743 (the unconditional Gemini fold plus the self-evident reasoning > completion fold on Chat Completions, no schema change) merges independently now. #744 waits for this PR and #743, then rebases onto `main` (GitHub retargets it when this branch merges). No action needed on this thread beyond merging when the platform's skew-tolerant reader is confirmed live.

SilenNaihin:
Train status for the coordinated release (relayed for the pin/tag owner): the incident train content is COMPLETE on main — #736 (schema v2) + #739 (per-alias fail-safe) + #740 (truncation hardening matrix) + #741 (identity stability, schema v3 + de-literaled authored schema_version) + #745 (provider_quota classing + health circuit). The platform pin-time checklist for the whole train lives in #745's PR body ("Platform pin-time requirements"). Two operational notes: (1) the prod fleet is running the #1156 skew-tolerance hotfix (verified live: v107 workers, zero digest errors, pointer converged), so the schema-3 identity move is safe by design — expect the one-time catalog republication churn documented in #741's body when the pin rolls; (2) failure_class is free text platform-side, so no migration is needed for the new provider_quota value. The tag can cut whenever the release owner is ready.

PR #744, opened 2026-09-03
gateway: declare additive-reasoning Chat Completions wires per deployment (reasoning_tokens_additive)

SilenNaihin:
## HELD: stacked on #741 (base branch `identity-stability`); do not merge before #741 and #743

This PR adds a field to `GatewayDeploymentCapabilities`. On `main`'s identity serializer (`canonical_json_bytes`, `exclude_none=False`, every default emitted) that moves every deployment digest in the catalog, which is exactly the 0.7.22 -> 0.7.25 incident. On #741's exclude-default serialization a default-valued field contributes zero identity bytes, so this lands digest-neutral with no `SNAPSHOT_SCHEMA_VERSION` bump. It is therefore based on `identity-stability`; GitHub retargets it to `main` when #741 merges. It also carries #743's commit (the fold itself) so the diff is coherent; that commit disappears from the diff once #743 is on `main` and this branch is rebased.

## What it does

`experientiallabs/platform#1157` step 1-3: per-deployment declaration that a Chat Completions wire reports reasoning OUTSIDE `completion_tokens`.

- `exp/common/models/catalog.py`: `GatewayDeploymentCapabilities.reasoning_tokens_additive: bool = False` (documented).
- `exp/runtime/gateway/native_execution.py::deployment_wire_entry`: carries the flag to the data plane next to `fireworks_reasoning_route_sha256`.
- Rust: `waterfall.rs::DeploymentWire.reasoning_tokens_additive` (`serde(default)`), threaded through `UpstreamRelay` into the `Normalizer` as a new `NormalizerOptions { reasoning_content_route_sha256, reasoning_tokens_additive }` (replaces the `new_with_reasoning_content_route` constructor variant rather than adding a third one); `openai_compatible_usage(value, declared_additive)` folds when declared OR when reasoning > completion (#743's self-evident case).
- Gemini needs no declaration (always additive, always folded in #743); Anthropic, Bedrock, and Responses publish subset semantics and ignore the flag.

## Identity evidence

On this branch (`SNAPSHOT_SCHEMA_VERSION = 3`, exclude-default identity):

```
default caps under identity serialization:  {}
declared caps under identity serialization: {'reasoning_tokens_additive': True}
```

`test_identity_digest_is_pinned_until_a_deliberate_schema_version_bump` and `test_added_defaulted_fields_and_explicit_defaults_do_not_perturb_identity` pass untouched with the new field, which is #741's stated contract for additive growth. Declaring the flag on a deployment changes that deployment's digest, as a catalog fact should.

On `main` today the same default `GatewayDeploymentCapabilities()` serializes to 770 bytes of identity (`{"maximum_stop_sequences":null,...,"supports_audio_input":false,...}`), so this field cannot land there without a bump; the bump is #741's.

## Heuristic vs declaration (why both)

#743's heuristic (fold when reasoning > completion) closed every observed prod row (platform#1157: 2/2 Grok rows, both live probes, and every additive receipt captured for this work had reasoning 5x to 200x the answer). It misses an additive provider only when reasoning is no larger than the visible answer, which for a reasoning model means a long answer after brief thinking. The declaration closes that case exactly and is what the platform already stamps on both Grok Azure lanes (`update_route` drafts in platform#1157, applied). `declared_additive_deployment_folds_reasoning_even_below_the_visible_answer` pins the difference: 30 in / 40 out / 5 reasoning -> declared 45 output, undeclared 40.

## Validation

- `cargo fmt --check`, `cargo clippy --no-default-features --all-targets -D warnings`, `cargo test --no-default-features`: 153 passed.
- `ruff check`, `ruff format --check`, `ty check`: clean.
- Targeted pytest (gateway_catalog, catalog, native_execution, dialect parity, native bridge, budgets, repo layout): 233 passed.
- Native crate 0.3.23 -> 0.3.24 (on top of #743's 0.3.23).

## Platform follow-ups (after the pin bump that carries this)

- Add `reasoning_tokens_additive` to `PROTOCOL_BOOLEAN_CAPABILITY_KEYS` (engine-pinned test) and to #947's dialect contract registry / the authored `azure_openai` Grok contract.
- The build warning in `explabs/gateway/catalog.py` (host lane declares the flag, pinned engine cannot fold) goes quiet.
- Observation, not changed here: `budgets.py::maximum_attempt_cost_micro_usd` bounds the reservation by the caller's output ceiling; on an additive wire the folded output can exceed `max_tokens` (xAI counts reasoning outside the cap), so the ceiling can under-reserve on declared lanes. Settlement is exact; only the pre-charge bound is affected.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/reference/gateway-architecture.md
+++ b/docs/reference/gateway-architecture.md
@@ -215,10 +215,11 @@ of `output_tokens` and `cached_input_tokens` a subset of `input_tokens`, and set
 subset at its own rate and the remainder at the base rate. Wires that report reasoning outside
 their output total are folded by the native usage mappers before the counts leave the data plane:
 Gemini `thoughtsTokenCount` is additive by Google's definition and always folds into
-`output_tokens`; on Chat Completions a `reasoning_tokens` count above `completion_tokens` is
-impossible under subset semantics, so it identifies an additive provider (xAI, natively or relayed
-by Azure Foundry) and folds, while OpenAI, OpenRouter, Fireworks, and DeepSeek pass through
-untouched. Anthropic and Bedrock bill thinking inside their output total and publish no separate
+`output_tokens`; on Chat Completions a deployment whose catalog declares
+`reasoning_tokens_additive` (xAI, natively or relayed by Azure Foundry) folds every reported
+reasoning count, and an undeclared deployment folds when `reasoning_tokens` exceeds
+`completion_tokens`, which is impossible under subset semantics and so identifies an additive
+provider on its own; OpenAI, OpenRouter, Fireworks, and DeepSeek pass through untouched. Anthropic and Bedrock bill thinking inside their output total and publish no separate
 count, so their reasoning subset stays unknown. The customer-visible `completion_tokens` and
 `total_tokens` therefore match what is billed on every lane.
 
--- a/docs/reference/gateway-architecture.md
+++ b/docs/reference/gateway-architecture.md
@@ -210,6 +210,18 @@ while keyed replay creates no new reservation. A period is the immutable UTC buc
 month. Management and remaining-allocation reports are CLI surfaces only. There is no budgets
 dashboard.
 
+Normalized usage follows OpenAI subset semantics on every wire: `reasoning_tokens` counts a subset
+of `output_tokens` and `cached_input_tokens` a subset of `input_tokens`, and settlement prices the
+subset at its own rate and the remainder at the base rate. Wires that report reasoning outside
+their output total are folded by the native usage mappers before the counts leave the data plane:
+Gemini `thoughtsTokenCount` is additive by Google's definition and always folds into
+`output_tokens`; on Chat Completions a `reasoning_tokens` count above `completion_tokens` is
+impossible under subset semantics, so it identifies an additive provider (xAI, natively or relayed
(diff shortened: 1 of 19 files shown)


SilenNaihin:
Status after #743's review revision: #743 now lets the provider's `total_tokens` decide the fold on both OpenAI-shaped wires (`input + output + reasoning` folds, `input + output` passes through), which closes the additive-but-smaller regime for every provider that publishes a total, xAI included. That removes the case this declaration was designed for. Recommendation: keep this PR held as a Draft only as an escape hatch for a future additive provider whose usage carries no informative total; if none appears, close it. This branch also carries the pre-review version of #743's commit and must be rebased onto `identity-stability` + `main` before any further work.

SilenNaihin:
Closing as speculative surface: #743's `fold_openai_shaped_reasoning` lets the provider's own `total_tokens` decide the fold on both OpenAI-shaped wires (`input + output + reasoning` folds, `input + output` passes through, magnitude fallback only without a decisive total), which covers xAI natively and via Azure Foundry on Chat Completions and Responses without any catalog declaration. A future additive provider whose usage carries no informative total is the only case this flag would serve; reopen from this branch (rebased) if one appears.

PR #745, opened 2026-09-03
gateway: class provider 402 as provider_quota deadness, never invalid_request

SilenNaihin:
## Incident

TokenHub (the Tencent relay) answered HTTP 402 with provider code 401008 — "free trial quota exhausted and postpaid billing is not enabled — enable in Console > Online Inference Service" — on **every** request shape. The gateway classed all of it `invalid_request` with the caller-facing text "provider rejected the request; verify the request fields against the model alias capabilities". 652 customer-visible errors blamed the caller's fields for the provider account's billing state, the rung never failed over (`invalid_request` is failover-ineligible), and diagnosis chased the capability/over-claim path for hours because `invalid_request` reads as a request-shape problem by contract.

## Change

**New failure class `provider_quota`** (`GatewayFailureClass.PROVIDER_QUOTA` / `FailureClass::ProviderQuota`): the provider ACCOUNT cannot pay (trial quota exhausted, billing disabled). Deliberately distinct from:

- `quota_exceeded` — that is the CALLER's gateway credit; reusing it would tell customers *they* are out of quota (`429 insufficient_quota`), misattributing in the opposite direction;
- `throttled` — a 402 is sticky operator-actionable deadness, and `throttled` is in the `maximize_cache` no-failover set, which would trap cache-mode pools on the unfunded rung surfacing 429s forever.

Behavior: `failover_eligible=True`, not same-deployment-retryable, fails over in **every** failover mode (deadness family), feeds the health circuit, and publicly renders through the provider-side fall-through (`502 all_routes_failed`, `api_error`) with the message "provider account quota or billing is exhausted; ask the gateway operator to fund or enable the provider account" — never a verify-your-fields 400.

HTTP 402 maps to it in **both mirrored seams**: the native data plane (`upstream.rs::transport_failure`) and the python taxonomy (`providers/errors.py::_transport_failure`). The mapping is dialect-agnostic (status-only), so anthropic/gemini/bedrock lanes get the same treatment.

## Regression fixtures

- Rust: a loopback-HTTP test drives `open_stream` against a canned 402 carrying the **literal TokenHub body** (code 401008) and asserts `ProviderQuota`, failover eligibility, content-free failure (the error body is never read for non-`InvalidRequest` classes), and that the message never says "request fields". The status parity table gains the 402 row.
- Python: both parametrized tables (class/flags and exact message) gain the 402 rows with the incident context inline.
- The Anthropic error-translation parity goldens gain the `provider_quota` entry (generated through the real renderer + native translator; additive-only diff).

## Chores

Native crate 0.3.22 → **0.3.24** (0.3.23 is reserved by in-flight #740) with pin/lock/wheel chores. `SNAPSHOT_SCHEMA_VERSION` untouched — failure classes are not catalog identity.

## Platform pin-time requirements (for the 0.7.28 coordinated-train pin PR)

This is the checklist for whoever authors the platform pin bump that ships this train (#736 + #739 + #740 + #741 + #745). Verified against the platform repo on 2026-09-03:

**From #745 (this PR):**
- `gateway_attempts` / `gateway_usage` `failure_class` columns are free text (`pg_catalog.text`; the only constraint anywhere is a 1..128 length check on `gateway_error_alert_rules.failure_class`) — **no migration needed**; the new `provider_quota` string just starts appearing in rows.
- Platform Python has **no exhaustive `GatewayFailureClass` matches** (only specific-member comparisons with fallthrough: `ledger.py` QUOTA_EXCEEDED/CANCELLED checks), and the enum member itself arrives via the engine pin — **no platform code change needed**.
- Optional operator step: add a `gateway_error_alert_rules` row pinning `failure_class = 'provider_quota'` so a future billing-dead house lane pages the operator while requests spill over silently (that alert is the whole point of the class).

**From #741 (identity stability, schema v3):**
- Hard precondition: platform #1156 (skew-tolerant `read_published_catalog_snapshot` + `authored_sha256`) must be **live in production before this train's pin deploys** (team-lead holds the signal).
- Expect a one-time catalog republication after the pin rolls: every digest re-derives under the exclude-default identity, so convergence republishes each alias under a new `catalog_sha256`. New usage/attempt rows key to the new digests; old rows keep the old ones (attribution is append-only, nothing to migrate).
- No platform code touches `SNAPSHOT_SCHEMA_VERSION` and no test pins a literal catalog digest (the one identity assertion in `catalog_test.py` recomputes both sides through the engine) — **no platform code change needed**.

## Validation

- Full `uv run pytest -q`: 3379 passed, 6 skipped
- `cargo test --no-default-features` (PYO3_PYTHON pinned): 144 passed incl. the new 402 tests; clippy/fmt clean

### DIFF
--- a/exp/runtime/gateway/contracts.py
+++ b/exp/runtime/gateway/contracts.py
@@ -917,6 +917,10 @@ class GatewayFailureClass(StrEnum):
     TIMEOUT = "timeout"
     PROVIDER_AUTHENTICATION = "provider_authentication"
     PROVIDER_NOT_FOUND = "provider_not_found"
+    # The provider ACCOUNT cannot pay for the request (trial quota exhausted,
+    # billing not enabled): operator-actionable deadness that fails over in
+    # every mode. Distinct from QUOTA_EXCEEDED, the CALLER's gateway credit.
+    PROVIDER_QUOTA = "provider_quota"
     REFUSAL = "refusal"
     MALFORMED_RESPONSE = "malformed_response"
     PROVIDER_INTERNAL = "provider_internal"
--- a/exp/runtime/gateway/health.py
+++ b/exp/runtime/gateway/health.py
@@ -14,6 +14,11 @@ DeploymentHealthKey = tuple[str, str, str]
 _HARD_FAILURES = {
     GatewayFailureClass.PROVIDER_AUTHENTICATION,
     GatewayFailureClass.PROVIDER_NOT_FOUND,
+    # An unfunded provider account (402) is the same sticky operator-actionable
+    # deadness as a bad credential: open the circuit immediately so the dead
+    # rung stops being attempted first, and let the half-open probe rediscover
+    # it once the operator funds or enables the account.
+    GatewayFailureClass.PROVIDER_QUOTA,
 }
 _OPERATIONAL_FAILURES = {
     GatewayFailureClass.TRANSPORT,
--- a/exp/runtime/gateway/health_test.py
+++ b/exp/runtime/gateway/health_test.py
@@ -68,6 +68,21 @@ def test_operational_failures_still_open_the_circuit() -> None:
         assert not registry.claim(_KEY)
 
 
+def test_provider_quota_opens_the_circuit_immediately_like_other_hard_deadness() -> None:
+    """An unfunded provider account (402) suppresses the rung after ONE failure.
+
+    Without this, the billing-dead deployment stays first-in-line and burns one
+    wasted attempt per request before every failover; with it, the circuit opens
+    at once and the half-open probe rediscovers the rung when the operator
+    funds or enables the account.
+    """
+    registry = DeploymentHealthRegistry(failure_threshold=2, clock=lambda: 100.0)
+
+    registry.failed(_KEY, _failure(GatewayFailureClass.PROVIDER_QUOTA))
+
+    assert not registry.claim(_KEY)
+
+
 def test_throttle_storms_still_suppress_the_deployment() -> None:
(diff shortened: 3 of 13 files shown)


SilenNaihin:
Addressed the P1: `PROVIDER_QUOTA` now sits in the health registry's `_HARD_FAILURES` beside `provider_authentication`/`provider_not_found`, so the circuit opens on the FIRST 402 and the billing-dead rung stops being attempted first on every request; the half-open probe rediscovers it once the operator funds or enables the account. Regression test: `test_provider_quota_opens_the_circuit_immediately_like_other_hard_deadness`. (Also rebased over the merged #740/#741 train, keeping crate 0.3.24 and restoring #740's h2 dev-dependency that the conflict resolution initially dropped — full suite 3384 green, cargo 148 green.)

SilenNaihin:
Train status for the coordinated release (relayed for the pin/tag owner): the incident train content is COMPLETE on main — #736 (schema v2) + #739 (per-alias fail-safe) + #740 (truncation hardening matrix) + #741 (identity stability, schema v3 + de-literaled authored schema_version) + #745 (provider_quota classing + health circuit). The platform pin-time checklist for the whole train lives in #745's PR body ("Platform pin-time requirements"). Two operational notes: (1) the prod fleet is running the #1156 skew-tolerance hotfix (verified live: v107 workers, zero digest errors, pointer converged), so the schema-3 identity move is safe by design — expect the one-time catalog republication churn documented in #741's body when the pin rolls; (2) failure_class is free text platform-side, so no migration is needed for the new provider_quota value. The tag can cut whenever the release owner is ready.

PR #750, opened 2026-09-03
gateway: srn sampling — drop+disclose+serve instead of rejecting (Task A)

SilenNaihin:
## What

A route that accepts `temperature`/`top_p` **only** at `reasoning_effort=none` (`sampling_requires_reasoning_none`, e.g. gpt-5.6-sol/luna) used to 400 ("temperature is not supported by this model route") when a caller sent `temperature`/`top_p` with reasoning on. The model **does** accept sampling — just not at that effort — so a hard reject is worse than adapting. Now the request **serves** with the sampling control dropped before dispatch and **disclosed** to the caller. (Design A, as approved.)

## How

- **`route_generation_parameter_requests`** (streaming_requests.py): split the srn-block from the genuinely-unsupported case. When every rung **declares** the control but at least one blocks it *solely* because it's a reasoning route not at `effort=none`, drop it via the existing `ignore()` seam (nulls the provider field + records the disclosure) instead of raising. A rung that never declares the control (Anthropic constrained `[1,1]`) still **hard-rejects** — nothing to honor at any effort. **No Rust/schema change** — the drop rides the existing `x-experiential-ignored-parameters` body key; the redundant payload-build guard (reasoning_compat.py) auto-no-ops once the provider field is nulled.
- **Disclosure token** (owner chose the visible-warning form; no OpenAI-wire warning field exists): self-explanatory `temperature->dropped(set_reasoning_effort_none)` / `top_p->dropped(...)` — field-name prefix (still parseable) + the existing `->` transition convention + the actionable fix.
- **`generation_route_compat.py`**: the per-rung selection probe treated my drop as "compatible," which would lose a rung that could *honor* temperature. Fixed: selection now **prefers rungs that honor the caller's sampling exactly** and only falls back to serve-with-drop rungs when none honor it — preserving intent when any rung can (this is the "adapt to preserve intent" principle applied to selection).
- **docs/reference/gateway-architecture.md** documents the srn drop + the honor-first selection fallback. (`/docs/errors` is a platform web route — that copy is platform-side, flagged to the lead.)

## Tests

- `test_srn_sampling_drops_and_discloses_instead_of_rejecting` — served; `provider_request.temperature/top_p` nulled; `public_request` keeps caller values + both disclosure tokens.
- `test_srn_sampling_is_honored_at_explicit_none_reasoning` — `reasoning_effort=none` honors sampling, no disclosure.
- `test_genuinely_unsupported_sampling_still_hard_rejects` — Anthropic constrained → 400 `unsupported_parameter`.
- `test_generation_parameter_selection_uses_each_default_for_sampling` — heterogeneous route still narrows to the honoring rung.
- `test_generation_parameter_selection_serves_with_drop_when_no_rung_honors` — pure-srn route stays serviceable (drop-serve, not rejected).
- 648 provider+protocol tests + gateway admission green; ruff/ty clean.

## Version

No bump — accumulating into the **held 0.7.28** release (matches #742/#743/#746). Pure Python, no native/schema change → P0-safe to develop; NO deploy until the reconciled release.

### DIFF
--- a/docs/reference/gateway-architecture.md
+++ b/docs/reference/gateway-architecture.md
@@ -305,8 +305,9 @@ idempotency header, so this surface never joins the keyed replay stores.
 Route admission preserves caller capabilities in three verbatim-preference layers before any
 coercion: operationally dead rungs are skipped (`dispatchable_route_profiles`), generation
 controls narrow the waterfall to the rungs that preserve every exact value
-(`compatible_generation_parameter_profile_indexes`), and each remaining deployment passes the
-capability preflight plus payload build. Only when zero rungs survive does the
+(`compatible_generation_parameter_profile_indexes`), falling back to the rungs that can serve the
+request only through a disclosed drop when no rung preserves it, and each remaining deployment
+passes the capability preflight plus payload build. Only when zero rungs survive does the
 capability-preservation policy (`exp/runtime/models/providers/capability_policy.py`) attempt one
 minimal COERCE-WITH-DISCLOSURE: a reasoning effort snaps to the nearest level any rung supports
 on the canonical ladder (ties prefer the lower level), ANY effort on a route with no reasoning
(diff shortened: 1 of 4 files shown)


SilenNaihin:
Folded into #752 (the combined request-adaptation framework PR) — the srn commit is the base of #752's stack, and everything builds on its honor-first selection. Closing to keep one combined PR per the owner's request.

PR #756, opened 2026-09-03
gateway/batch: batch support for OpenAI-compatible providers (Doubleword)

### LINKED ISSUE #755: Batch API support for OpenAI-compatible providers (Doubleword)
## Summary

The `/v1/batches` lane ships batch clients only for OpenAI, Anthropic, and OpenRouter, each pinned to the vendor's own host. There is no client for a custom OpenAI-compatible provider, so a provider such as Doubleword cannot be reached through the Batch API even though it serves the OpenAI Batch dialect verbatim.

## Proposal

Add a batch provider client for OpenAI-compatible endpoints, so any provider that serves the OpenAI Batch API (file upload plus `/batches` create, poll, cancel, and file-id results) can be registered by binding its base URL. Doubleword is the first such provider.

## Scope

Client and registration only. Enabling the lane for a host (the `BatchControlPlane`, the catalog row, and the secret resolver) stays the host's responsibility, as it is for the existing providers. No engine, plane, route, or contract changes.

## Verification

Doubleword's batch API is a verbatim OpenAI dialect, and small batches complete in about two minutes, so a gated live test can exercise the real endpoint end to end.

jaedmunt:
Closes #755

## What

Adds a batch provider client so Doubleword, and any OpenAI-compatible endpoint, can be reached through the `/v1/batches` lane. Doubleword speaks the OpenAI Batch dialect verbatim (file upload, `/batches` create, poll, cancel, file-id results), so the client reuses the OpenAI batch path and only binds the base URL.

## Two shapes, as separate commits

The existing batch clients are one concrete class per vendor, so this offers both approaches and you can pick the one you prefer:

1. Concrete `DoublewordBatchClient`, a small subclass overriding the provider name and base URL. Matches the current per-vendor pattern.
2. Generic `OpenAICompatibleBatchClient`, taking a configurable base URL and provider, with `DoublewordBatchClient` rebased onto it. Reusable for any OpenAI-compatible provider.

Drop the second commit for the concrete-only version, or keep both for the generic seam. Either way the registration and behaviour are identical.

## Scope

Client and registration only. Enabling the lane for a host (the `BatchControlPlane`, the catalog row, and the secret resolver) stays the host's job, as for the existing providers. No engine, plane, route, or contract changes.

## Testing

- Wire-level unit tests over `httpx.MockTransport`: the client drives the OpenAI batch dialect at Doubleword's host, and the generic client targets its configured base.
- Gated live tests (`EXP_LIVE_BATCH=1` plus `DOUBLEWORD_API_KEY`): a real two-line batch through the direct client, and the full served `/v1/batches` lane over real HTTP against Doubleword's batch API. Both pass, and small batches settle in about thirty seconds.
- `ruff` and `ty` clean; batch package suite green.

## Open question

`OpenAICompatibleBatchClient` inherits from the vendor-named `OpenAIBatchClient`. If you would rather not couple the generic client to the OpenAI class, a shared `_OpenAIDialectBatchClient` base for both is an easy alternative. Draft while you weigh the client shape and that inheritance.

### DIFF
--- a/exp/runtime/gateway/batch/providers.py
+++ b/exp/runtime/gateway/batch/providers.py
@@ -702,8 +702,16 @@ class OpenRouterBatchClient:
         )
 
 
+class DoublewordBatchClient(OpenAIBatchClient):
+    """Doubleword Batch API: the OpenAI batch dialect at Doubleword's endpoint."""
+
+    provider = "doubleword"
+    _base = "https://api.doubleword.ai/v1"
+
+
 PROVIDER_CLIENTS: dict[str, ProviderBatchClient] = {
     OpenAIBatchClient.provider: OpenAIBatchClient(),
     AnthropicBatchClient.provider: AnthropicBatchClient(),
     OpenRouterBatchClient.provider: OpenRouterBatchClient(),
+    DoublewordBatchClient.provider: DoublewordBatchClient(),
 }
--- a/exp/runtime/gateway/batch/providers.py
+++ b/exp/runtime/gateway/batch/providers.py
@@ -702,11 +702,36 @@ class OpenRouterBatchClient:
         )
 
 
-class DoublewordBatchClient(OpenAIBatchClient):
+class OpenAICompatibleBatchClient(OpenAIBatchClient):
+    """The OpenAI batch dialect at a configurable base URL and provider name.
+
+    Any provider that serves the OpenAI Batch API verbatim (file upload plus
+    ``/batches`` create, poll, cancel, and file-id results) is reachable by
+    binding its ``base_url`` and provider identity on the instance.
+    """
+
+    def __init__(
+        self,
+        *,
+        provider: str,
+        base_url: str,
+        transport: httpx.AsyncBaseTransport | None = None,
+    ) -> None:
+        """Bind the provider identity and endpoint for this instance."""
+        super().__init__(transport=transport)
+        self.provider = provider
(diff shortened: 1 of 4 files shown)


SilenNaihin:
Related finding from the platform host (2026-09-03), in case it shapes the client seam here: OpenRouter's beta batch API takes the PLAIN slug (the `:batch` listing id is a 400: "Model 'openai/gpt-4.1-nano:batch' does not have a :batch endpoint."), and even with the plain slug it refuses most OpenAI-author and Anthropic-author models (19 of 31 openai twins, 10 of 12 anthropic twins in the platform catalog; google/deepseek/qwen/z-ai/mistral/moonshot/minimax/meta/thinkingmachines all accept). The platform is therefore routing OpenAI-author batch twins to the direct `OpenAIBatchClient` (api.openai.com files + batches) and Anthropic twins to `AnthropicBatchClient`, with OpenRouter only for the providers it can batch, so a generic OpenAI-dialect client bound to a configurable base URL (your commit 2 shape) is the seam the host would reach for next for any further OpenAI-compatible batch provider. No duplicate PR from our side. Also landed today: #760 (a failed batch names the provider's `error.message` on the batch object; that 400 body was previously dropped at `_checked`) and #762 (uniform 401 on an invalid key across the batch routes).

jaedmunt:
Hey, are there any changes necessary our side for this? :)

PR #759, opened 2026-09-03
gateway: hash and index each distinct catalog object once in route resolution

SilenNaihin:
## Problem (routed from the platform's catalog-incident follow-ups, P2)

`CatalogRouteResolver._index_catalogs` verified `catalog.identity_sha256()` per alias KEY, but a repoint mints every key against ONE immutable `NormalizedGatewayCatalog` object. Measured on the real production document: one state build re-hashed the same 6.5 MB catalog 732 times — 35.4 s at 48 ms/call, the dominant term of the 51 s `build_catalog_state` cost and of the 11-minute pod warms seen during the 09-03 deploy rollbacks. Each key also rebuilt its own `_CatalogView` (two dicts of ~2,400 entries at production scale): another ~0.3–1 s and 732 duplicate dict pairs held in memory.

## Change (strictly resolver-internal)

The digest is computed **once per distinct catalog object** and compared **per key**; the frozen `_CatalogView` is likewise built once per object and shared across its keys. Corruption semantics are unchanged and regression-pinned:

- a same-version object pinned under a second, WRONG digest still fails closed (the per-key comparison survives the memoization);
- the foreign-snapshot tolerance path is untouched;
- no digest computation or serialization changes anywhere — this is why the resolver-internal dedupe was chosen over memoizing `identity_sha256` on the frozen model itself (option (a)): it cannot interact with the identity contract at all, and `_index_catalogs` is the only seam with the 732-to-1 key-to-object fan-in.

`test_index_hashes_a_shared_catalog_object_once_and_shares_its_view` pins all three properties (one hash for 50 shared keys, shared view identity, wrong-second-digest still raises).

## Platform note

The platform ships a temporary identity-caching `NormalizedGatewayCatalog` subclass for the same hot path (platform #1168); it becomes removable at the engine pin bump that carries this change.

## Validation

- Full `uv run pytest -q`: 3390 passed, 6 skipped
- No `native/` changes; no crate bump

### DIFF
--- a/exp/runtime/gateway/routing.py
+++ b/exp/runtime/gateway/routing.py
@@ -454,17 +454,36 @@ def _index_catalogs(
         ValueError: A same-version catalog does not match its declared digest.
     """
     indexed: dict[tuple[str, str], _CatalogView] = {}
+    # A repoint mints every alias key against ONE immutable catalog object
+    # (hundreds of keys per snapshot in production), and identity_sha256
+    # re-hashes the whole multi-megabyte document, so the digest is computed
+    # once per distinct object and compared per key; the frozen view is built
+    # and shared once per object too. Per-key hashing made one state build
+    # re-hash the same 6.5 MB catalog 732 times (~35 s of a ~51 s build).
+    # Object ids are stable here because ``catalogs`` keeps every catalog
+    # alive for the whole loop.
+    identity_by_object: dict[int, str] = {}
+    view_by_object: dict[int, _CatalogView] = {}
     for key, catalog in catalogs.items():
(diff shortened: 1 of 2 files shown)


SilenNaihin:
Merged. Bookkeeping for the platform pin that first carries this change: platform #1168's interim identity-cache subclass becomes dead weight — delete `_IdentityCachedNormalizedCatalog` + `_identity_cached` (both normalize call sites in `explabs/gateway/catalog.py`); the counting regression test can stay pointed at the engine behavior. Filed as a platform follow-up item.

PR #761, opened 2026-09-03
fix(openai_protocol): accept store:false on Chat Completions (no-op retention opt-out)

phoenix-server:
## Problem

OpenAI-compatible agents (`omp`, `opencode`, `pi`) hardcode `store: false` on every Chat Completions request to opt out of provider-side retention. The gateway's Chat manifest declares `store` UNSUPPORTED, so these agents get a hard 400 before dispatch:

```
The parameter 'store' is not supported by this gateway profile. Remove the field and resend the request.
```

Reported against a fresh agent fleet (omp 18.1.3 against an experiential gateway): the agent could not produce a single reply.

## Root cause

`store` is genuinely supported on the **Responses** surface (it feeds `response_store`), but the **Chat** surface lists it under `UNSUPPORTED` in `CHAT_MANIFEST`, so `_validate_manifest` rejects it pre-dispatch. The gateway never retains Chat output on any rung, so `store: false`, the only value these agents send, is already satisfied.

## Fix

Mirror the existing `n`-field precedent (accepted only at its no-op default; any other value is a named rejection):

- `manifest.py`: declare `store` SUPPORTED in `CHAT_MANIFEST` (remove it from the UNSUPPORTED list), with a rationale.
- `wire_models.py`: add `store: bool | None` to `_ChatRequest` with a validator that accepts only `false`/absent and rejects `true`. A `store: true` request asks the gateway to retain output for distillation/evals, which it never does. Per the codebase's anti-silent-drop policy, that must be a named rejection, not a silent no-op.

## Tests

- `test_chat_decoder_accepts_store_false_opt_out`: the exact omp/opencode request shape now decodes.
- `test_chat_decoder_rejects_store_true_retention_request`: `store: true` stays a named `invalid_parameter` on `store`.
- `manifest_test`: chat `store` is now `SUPPORTED`.

Verified: the real 400-ing body (`store:false` plus `stream`, `stream_options`, `max_completion_tokens`, `reasoning_effort`) decodes cleanly. `requests_test.py` and `manifest_test.py` pass (163 tests).

### DIFF
--- a/exp/runtime/openai_protocol/manifest.py
+++ b/exp/runtime/openai_protocol/manifest.py
@@ -65,6 +65,13 @@ CHAT_MANIFEST = CompatibilityManifest(
         # Accepted only at its no-op default of 1 (the wire model enforces
         # the value): Copilot hardcodes n:1 on every Chat request.
         _field("n", CompatibilityDisposition.SUPPORTED),
+        # Retention request accepted only at its no-op default of false (the
+        # wire model enforces the value): OpenAI-style agents (omp, opencode,
+        # pi) hardcode store:false on every Chat request to opt out of
+        # provider-side retention. This gateway never retains Chat output on
+        # any rung, so false is already satisfied and store:true is rejected:
+        # silently dropping a retention request would be dishonest.
+        _field("store", CompatibilityDisposition.SUPPORTED),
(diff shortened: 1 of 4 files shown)


SilenNaihin:
Thanks for the PR @phoenix-server

PR #774, opened 2026-09-04
gateway: carry tool_result image blocks; degrade with disclosure off-route; name union misses

SilenNaihin:
## Incident (owner-reported, wedged live Claude Code session)

The Messages-surface wire model rejected image sub-blocks inside `tool_result.content` — real Anthropic wire that Claude Code's Read-on-image and computer-use tools emit routinely — and the pydantic union leaked the misleading error `Invalid value for 'messages.N.content.str': Input should be a valid string`, naming the union's string arm when the actual problem was an unsupported sub-block in the list arm. The block is baked into conversation history, so the 400 wedged every later turn of the session, and the misleading error sent the diagnosis chain toward the caller's request shape.

## Fix 1: accept, then route by capability

- `_ToolResultBlock.content` accepts `image` sub-blocks; they decode into the tool message's `content_parts` (the canonical model now allows text+image parts on `role="tool"`; text-only results keep the exact prior serialization and digests, regression-pinned).
- **All-Anthropic, image-capable route:** verbatim pass-through — `anthropic_blocks` re-emits the caller's exact block run (ordered text/image blocks, cache markers preserved via the shared media helpers).
- **Route with any non-Anthropic rung:** no other wire defines a tool-result image carrier, so route shaping substitutes positional placeholder text (`[image omitted: this model route cannot carry tool-result images]`) with an `ignored_parameters` disclosure (`messages.content.tool_result.image->placeholder`), mirroring the `context_management` precedent — never a silent drop at encoding, never a 400.
- **All-Anthropic, NON-vision route** (fable-5.1's current route): per-rung preflight rejects `image_input` as before, and the route-wide rejection now coerces to the same disclosed placeholder degrade — but ONLY when every image in the request lives in a tool message. Top-level user images keep the fail-closed contract (the caller can re-send those; a history block they cannot remove).

## Fix 2: union-miss error rendering

- The deepest-location pydantic error is rendered instead of the first arm's, so a list-shaped problem never reports `content.str`; CamelCase arm labels and scalar type spellings are stripped from public paths.
- The pre-scan hint names the exact offending block path, and an unknown `tool_result` sub-block type gets its own message: `messages.N.content.i.content.j: unsupported block type 'X' inside tool_result content; only text and image sub-blocks are supported.`

## Regression fixtures (as ordered)

The exact repro (user text; assistant `tool_use`; user `tool_result` whose content is one base64 1×1 PNG) is decoded through the real surface decoder and driven through `admitted_route_requests` × (vision-capable route → verbatim pass-through, non-vision route → disclosed placeholder + coercion accounting) × (streamed, non-streamed); plus decode/round-trip/coercion/validator unit fixtures and the never-`content.str` error pins.

## Notes

- Python-only: no `native/` changes, no crate bump. Rides the next train.
- **Inherited red:** the `gate` check is failing on main itself at bc8c1566 — #772 pushed `exp/runtime/gateway/native_bridge.py` to 1030 lines, tripping the 999-line repo-layout test. This PR's full suite is green except that pre-existing failure (3563 passed, 1 failed = the layout test on an untouched file).

## Validation

- `uv run pytest -q`: 3563 passed, 6 skipped, 1 failed (the inherited main-red layout test above)

### DIFF
--- a/exp/common/models/content.py
+++ b/exp/common/models/content.py
@@ -297,6 +297,12 @@ class TextContentPart(ContractModel):
 
     kind: Literal["text"] = "text"
     text: str
+    cache_control: JsonObject | None = Field(default=None, exclude=True)
+    """Prompt-cache breakpoint the caller placed on this text block, re-emitted
+    verbatim when the block run itself re-emits (a tool_result's inner blocks
+    on Anthropic rungs) and dropped elsewhere: a cache hint changes cost, not
+    semantics. Like the other cache carriers it joins neither serialization
+    nor replay identity."""
 
 
 class ImageContentPart(ContractModel):
--- a/exp/runtime/anthropic_protocol/requests.py
+++ b/exp/runtime/anthropic_protocol/requests.py
@@ -12,11 +12,13 @@ everywhere and carried on the surfaces the Anthropic wire caches natively
 top-level automatic marker), and dropped on wires that do not cache a marked
 block because a cache hint changes cost, not semantics; ``image`` and PDF
 ``document`` blocks are retained as canonical content parts so a route that
-declares the matching input capability carries them, while a document
-inside ``tool_result`` content is rejected loudly because the serving
-surface cannot preserve it there; ``video`` blocks are rejected because the
-wire defines none. Unknown or unsupported fields are rejected with a
-field-specific error, never silently dropped. Errors raise
+declares the matching input capability carries them — including ``image``
+sub-blocks inside ``tool_result`` content (tool screenshots), which ride the
+tool message's content parts; a document inside ``tool_result`` content is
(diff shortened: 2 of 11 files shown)


SilenNaihin:
Addressed the P2: `TextContentPart` gains the same excluded verbatim cache-marker carrier `ImageContentPart` already has; the decode carries an inner text block's `cache_control` into the part and the Anthropic tool encoder re-emits it with the block run, so a Claude Code breakpoint landing on a text sub-block inside an image-bearing tool_result round-trips losslessly (regression: `test_a_cache_marker_on_a_tool_result_text_block_round_trips`, whole-block-run assertion). Text-only tool_results are unchanged (flattened content, marker on the tool_result block itself, as before).

PR #797, opened 2026-09-05
gateway: round-trip SDK 3.0 caller + list tool outputs; serve id-only reasoning replays (native 0.3.33)

### LINKED ISSUE #795: messages surface: unsupported reasoning-effort 400 names 'reasoning_effort', defeating Claude Code's output_config.effort auto-recovery
## Summary

On the Anthropic **messages** surface, a request carrying `output_config: {"effort": ...}` against a route whose profile doesn't allow that effort is rejected with an error that names the *translated OpenAI param*, not the field the client sent:

```
400 Reasoning effort 'high' is not supported by this model route. Supported values: 'medium'. (param: reasoning_effort)
```

Claude Code (2.1.261) sends `output_config.effort` on **every** request and has built-in recovery: on a 400 it latches "effort unsupported" and retries without the field — but only when the error matches one of its predicates (message contains `"effort parameter"` + `"not support"`, or `output_config`+`effort`+`"extra inputs are not permitted"`, or an `effort_unsupported` code). The current gateway phrasing matches none of them, so instead of degrading gracefully, every Claude Code turn hard-fails on affected routes (e.g. `hermes-4-405b`, whose live profile allows only `'medium'` while Claude Code defaults to `'high'`).

## Repro

```bash
curl -s $GW/v1/messages -H "x-api-key: $KEY" -H "anthropic-version: 2023-06-01" \
  -H "content-type: application/json" \
  -d '{"model":"hermes-4-405b","max_tokens":64,"output_config":{"effort":"high"},
       "messages":[{"role":"user","content":"hi"}]}'
# -> 400 ... (param: reasoning_effort)   <- client never sent that param
```

## Empirical proof of the client latch

Against a mock that returns the same 400 with the message
`"The effort parameter is not supported by this model route. Remove output_config.effort and resend the request. (param: output_config.effort)"`,
Claude Code logs `[effort] model ... rejected output_config.effort; latching unsupported and retrying without it.` (telemetry `tengu_effort_unsupported_retry`), retries **without** `output_config`, and the session proceeds normally. With the current gateway phrasing it does not retry and the turn dies.

## Proposed minimal fix

1. `exp/runtime/models/providers/streaming_requests.py:204` — make `effort_path` surface-aware for MESSAGES:
   ```python
   effort_path = (
       "reasoning.effort" if request.surface.value == "responses"
       else "output_config.effort" if request.surface.value == "messages"
       else "reasoning_effort"
   )
   ```
2. `exp/runtime/models/providers/errors.py` (`UnsupportedReasoningEffortError`) — keep the supported-values detail but include a client-recoverable phrase, e.g.:
   `"The effort parameter is not supported at this value by this model route. Supported values: 'medium'."`
   (contains `effort parameter` + `not support`, which triggers Claude Code's drop-and-retry).

Note `exp/runtime/models/providers/reasoning_compat.py` also hardcodes `param="reasoning_effort"` at its raise site.

## Impact

This is the last blocker-class 400 for Claude Code on reasoning-constrained non-Anthropic routes *after* #787 lands (thinking translation + is_error fold): CC sends `output_config.effort` independently of `thinking`, so #787 alone doesn't cover it. With this fix, CC degrades to effort-less requests on such routes with no client-side env workarounds — which on the hermes-4 openrouter lane also avoids enabling upstream reasoning that comes back as inline CoT in `content` (`reasoning_tokens: 0`, not separated by the provider).

Found while validating Hermes models in Claude Code through the gateway (2026-09-05, engine 0.7.35 era prod).

SilenNaihin:
## What this is

The follow-up wedge-fix train from the Sep 2026 adversarial stress test (the round-3 open findings), plus #795. Each item is a field or shape that is VALID to the upstream provider but that this gateway 400-killed on input or mangled on output; every one is baked into the caller's conversation history, so the rejection wedged whole sessions. All five now serve, following the drift-tolerance philosophy and the #784/#774/#787 precedents they extend.

### 1. SDK 3.0 `caller` round-trip (HIGH — the full #784 treatment)

`caller` is the SDK 3.0 programmatic tool-calling attribution (`{"type": "direct"}` / `{"type": "program", "caller_id": ...}`) on `function_call`, `function_call_output`, and `custom_tool_call` items. Before: rejected on input by the manifest's conscious-reject list AND stripped from provider output by the Rust encoder. Now it survives the whole loop, exactly like the #784 namespace: wire models + manifest accept-list on decode (validated only as an object — its internal shape is the provider's evolving surface), canonical carriers (`ToolCall.provider_caller`, `GatewayMessage.provider_tool_caller` with the tool-role validator), the outbound Responses payload rebuild, continuation retention (Rust + `remember_turn`), the Rust dialect parser (with completion identity checks and a malformed non-object 400), the client SSE encoder and non-streamed body, and replay identity — joining only when present so caller-free digests are byte-identical. Native crate 0.3.32 → **0.3.33** (Cargo.toml + crate pyproject + root pin + locks).

### 2. `function_call_output.output` list form (HIGH — the #774 tool_result treatment)

The SDK types `output` as `str | list of content parts` (tools that return images beside text). The list form now decodes onto the canonical tool message — text parts concatenate into `content`, image parts become `content_parts` (role=tool carries text+image since #774) — and re-emits as the typed part list on the native Responses wire (same image encoder as user messages; a failed invocation folds its `[tool error]` prefix in as a leading text part, derived from the canonical flag so replays never accumulate). An all-text list flattens back to the exact plain-string wire shape, so pre-list histories and digests are unchanged. An `input_file` part is a named 400: the canonical tool-message contract is text+image only (deliberate scope; widening to documents is a separate contract change). `custom_tool_call_output.output` was already opaque `JsonValue` carried verbatim — now pinned by fixture. Route shaping: the tool-result image keep-rule extends from all-Anthropic to all-Anthropic OR all-native-Responses routes (the Responses wire has its own carrier now); mixed/chat routes keep the #774 disclosed placeholder degrade, and non-vision rungs keep the preflight + coerce path.

### 3. Reasoning items without `encrypted_content` (MEDIUM — the hosted-item posture)

The SDK marks `encrypted_content` optional: a `store: true` flow replays reasoning by item id alone and the provider resolves it from stored state. An id-only reasoning item now carries VERBATIM as a native item to homogeneous native Responses routes — admission already restricts native items to exactly the one wire that can resolve the id — and the provider judges resolvability with its own named error, instead of this gateway 400ing SDK-legal input. An EMPTY `encrypted_content` stays a named 400 (malformed, not merely absent). **LIVE-VERIFIED post-review (session lead, house key, 2026-09-05), results verbatim:** created a stored gpt-5.1 response with reasoning effort low → output carried a reasoning item `rs_0ae4b217...`; replayed that reasoning item BY ID with NO `encrypted_content` in the input array alongside a new user message — (A) `store: true`: 200, status completed (`resp_...5b476...`); (B) `store: false`: ALSO 200, completed (`resp_...5c6f5c...`). OpenAI accepts encrypted-content-less reasoning items by id in BOTH store modes, so the verbatim-native-item posture is provider-confirmed, not just SDK-justified — and the fail-closed 400 shipped before this PR rejected a shape the provider demonstrably serves. (The wire-model docstring's "not verified live" caveat predates this probe; it gets corrected on the release-train PR.) The posture fails open to the provider's own verdict, mirroring the #792 hosted-item philosophy.

### 4. Attribution drop disclosure off the Responses wire (LOW)

A namespaced or caller-attributed call replayed through a chat/Anthropic/Gemini/Bedrock rung is rebuilt without its attribution; previously a silent drop. Route shaping now discloses each drop (`messages.tool_calls.namespace->dropped(unsupported_by_provider)`, `.caller->...`, `messages.tool_results.attribution->...`) — the call itself always survives with its exact name and arguments, so this is disclosure, never rejection.

### 5. #795: effort 400s match Claude Code's recovery latch

Claude Code sends `output_config.effort` on every request and auto-recovers from an effort rejection (drops the field, retries, `tengu_effort_unsupported_retry`) only when the 400 matches its predicates. The gateway named the translated internal param (`reasoning_effort`) with unmatched phrasing, so every turn on an effort-constrained route (e.g. hermes-4-405b, `medium`-only) hard-failed. Now: the Messages surface names `output_config.effort` and `UnsupportedReasoningEffortError` phrases the miss as "The effort parameter value 'high' is not supported by this model route. Supported values: 'medium'." — containing the exact "effort parameter" + "not supported" predicate. Closes #795.

## Fixtures

- Repo: caller round-trip (decode → carriers → payload rebuild, caller-free byte-identity, malformed named 400s, retention replay, replay identity binding, tool-role validator), list outputs (text+image mapping + typed re-emission, all-text flattening, input_file named 400, custom verbatim), id-only reasoning (native-item decode; empty-payload 400 repinned), keep-rule (all-Responses keeps, Responses+chat degrades), attribution disclosures (chat rung discloses, homogeneous Responses does not), effort latch (param + both phrasing predicates). Rust: normalizer caller round-trip + changed-at-completion + non-object 400, SSE/body re-emission, retention payload.
- Wedge-stress harness (`exp-wedge-stress/stress/run_attacks4.py`, out of repo): 14/14 ok against the real served engine, including the provider-EMITS-caller round trip over two generations, streamed and non-streamed.
- Three pre-existing tests pinned the old rejections and were updated to pin the new contract (id-only reasoning decodes; empty `encrypted_content` still 400s; an object `caller` accepted, a non-object one still names `input.N.caller`; the effort phrasing pin).

## Gates

- `uv run ruff format/check`, `ty check` clean; full `uv run pytest -q` green (3702 passed; the two w16 evidence tests require the clean committed checkout and pass on it).
- native/ touched: `cargo fmt --check`, `cargo clippy --no-default-features -D warnings`, `cargo test --no-default-features` (191) all green; crate version moved 0.3.32 → 0.3.33 as the gate requires; native parity tests green against the rebuilt wheel.

## Release train

Rides the next `exp-gateway-native` release with the #784/#786/#787 train: publish `experiential` + `exp-gateway-native` 0.3.33 wheels, then bump the platform's exact `==X.Y.Z` pins. Until then these fixes serve from-source checkouts only.

### DIFF
--- a/docs/reference/gateway-architecture.md
+++ b/docs/reference/gateway-architecture.md
@@ -178,6 +178,18 @@ their native event names, including continuation retention. The item-level `name
 `custom_tool_call` namespace) round-trips verbatim through decode, the client stream, and
 continuation retention: the provider rejects a namespaced call replayed without it, so the
 field joins replay identity when present and absent items keep their exact pre-existing shape.
+The SDK 3.0 programmatic tool-calling `caller` object on `function_call`,
+`function_call_output`, and `custom_tool_call` gets the same verbatim round trip (validated only
+as an object; its internal shape is the provider's). `function_call_output.output` accepts the
+SDK list form: text and image parts map onto the canonical tool message and re-emit typed, an
+all-text list keeps the plain-string wire shape, and any other part kind is a named 400. A
+reasoning input item without `encrypted_content` (a `store: true` replay by item id) carries
+verbatim to homogeneous native Responses routes and the provider judges resolvability. Off the
+native Responses wire, tool-call and tool-result attribution (`namespace`/`caller`/output
+`name`) drops with per-field disclosure — the call itself always survives — and a
+Messages-surface effort the route cannot serve rejects as `output_config.effort` with
+"effort parameter … not supported" phrasing, the exact predicate Claude Code's built-in
+drop-and-retry recovery latches on.
 
 Provider client-errors stay sanitized: no provider error prose or body content ever reaches the
 caller. The one provider-derived fact a 4xx rejection may relay is the parameter path the provider
--- a/exp/common/models/model.py
+++ b/exp/common/models/model.py
@@ -238,6 +238,18 @@ class ToolCall(ContractModel):
     from serialization like the other replay fields, and joins gateway replay
     identity explicitly.
     """
+    provider_caller: JsonObject | None = Field(default=None, exclude=True)
+    """Opaque SDK 3.0 ``caller`` attribution on a Responses tool-call item.
+
+    Programmatic tool calling attributes a ``function_call`` or
+    ``custom_tool_call`` to the program that invoked it (for example
+    ``{"type": "program", "id": ...}``). The object's internal shape is an
+    evolving provider surface, so it is validated only as an object and
+    round-trips verbatim like ``provider_namespace``: set when a Responses
+    caller replays an item carrying ``caller`` or when the provider emits
+    one. Excluded from serialization like the other replay fields, and joins
+    gateway replay identity explicitly.
+    """
 
     @model_validator(mode="after")
     def _require_matching_raw_arguments(self) -> ToolCall:
--- a/exp/runtime/gateway/contracts.py
+++ b/exp/runtime/gateway/contracts.py
@@ -247,6 +247,17 @@ class GatewayMessage(ContractModel):
     excluded from serialization like the other replay carriers, joining
     replay identity explicitly through :func:`canonical_request_sha256`.
     """
+    provider_tool_caller: JsonObject | None = Field(default=None, exclude=True)
+    """Opaque SDK 3.0 ``caller`` attribution on a ``function_call_output``.
+
+    Programmatic tool calling attributes the result to the program that
+    invoked the call; the object's internal shape is an evolving provider
+    surface, so it is validated only as an object and re-emitted verbatim on
+    the rebuilt item (mirroring ``ToolCall.provider_caller`` on the call
+    side). Excluded from serialization like the other replay carriers,
+    joining replay identity explicitly through
+    :func:`canonical_request_sha256`.
+    """
     provider_native_item: JsonObject | None = Field(default=None, exclude=True)
     """One verbatim OpenAI Responses input item the gateway carries opaquely.
 
@@ -372,7 +383,9 @@ class GatewayMessage(ContractModel):
         if self.role != "tool" and self.tool_call_id is not None:
             raise ValueError("tool_call_id is valid only for tool messages")
         if self.role != "tool" and (
-            self.provider_tool_name is not None or self.provider_tool_namespace is not None
+            self.provider_tool_name is not None
+            or self.provider_tool_namespace is not None
+            or self.provider_tool_caller is not None
         ):
             raise ValueError("tool-result attribution is valid only for tool messages")
         if self.role != "tool" and self.tool_is_error:
--- a/exp/runtime/gateway/contracts_test.py
+++ b/exp/runtime/gateway/contracts_test.py
@@ -1071,3 +1071,52 @@ def test_replay_identity_binds_tool_output_attribution() -> None:
     assert canonical_request_sha256(request(namespace="collaboration")) != canonical_request_sha256(
         plain
     )
+
+
+def test_replay_identity_binds_the_function_call_caller() -> None:
+    """A retained SDK 3.0 caller joins replay identity only when present."""
+    from exp.runtime.gateway.replay_identity import canonical_request_sha256
+
+    def request(*, caller: JsonObject | None) -> GatewayRequest:
+        return GatewayRequest(
+            surface=GatewayApiSurface.RESPONSES,
+            messages=(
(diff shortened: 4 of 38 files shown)


SilenNaihin:
Merged. This joins the #784+#786+#787 wedge-fix train for the next release: publish experiential + exp-gateway-native 0.3.33 wheels, then bump the platform's exact ==X.Y.Z pins (checklist in #787's Release train section). Until the release, all fixes serve from-source checkouts only. Regression net: exp-wedge-stress/stress/run_attacks4.py (14/14 ok against this build). Closes #795.

SilenNaihin:
Live-provider verification for item 3 (id-only reasoning replay), closing the docstring's "unverifiable from this environment" caveat: ran the probe against api.openai.com directly (2026-09-05, gpt-5.1, reasoning effort low). Created a stored response carrying a reasoning item (resp_0ae4b217…, rs_0ae4b217…), then replayed that reasoning item BY ID with NO encrypted_content alongside a new user message. Result: HTTP 200, status "completed", in BOTH store=true and store=false replay modes. The verbatim-native-item posture is provider-confirmed, not just SDK-justified — the fail-closed 400 this PR removes was rejecting a shape the provider demonstrably serves.

PR #809, opened 2026-09-05
Carry hosted request context through native admission and replay

kfallah:
The hosted gateway needs the caller's verified location during route admission, including before returning a cached response. Native callbacks currently lose this transport context, and the replay probe swallows route-policy 403s.

Carry a bounded, opaque `X-Exp-Request-Context` through HTTP and Responses WebSocket callbacks to an optional host `call_with_context` hook. Missing or ambiguous context stays unknown. Propagate typed policy 403s before replay lookup. The engine does not authenticate the header; hosted ingress must overwrite public copies and keep the listener private. Ordinary local control planes retain their callback API. The shipped type stub exposes the capability flag and accepts context-hook-only hosts.

Prepare Experiential 0.7.38 and native 0.3.36 for the dependent Platform country policy. No release is published by this PR.

Validation: 134 Python bridge and served-transport tests; 207 Rust tests; full repository Ruff check, format check and ty check. Served native tests cover discovery, HTTP inference surfaces, batches, idempotent claims and multiple WebSocket frames. Concurrent callback tests prove context isolation and missing-context behavior.

### DIFF
--- a/docs/reference/gateway-architecture.md
+++ b/docs/reference/gateway-architecture.md
@@ -445,3 +445,21 @@ Azure, and OpenRouter share compatible-stream coverage; Gemini and Bedrock have 
 fixtures. Live provider cells remain explicitly `not_run_requires_credentials` until a separately
 authorized run supplies dated evidence. Deterministic fixtures do not imply hosted-provider
 availability, billing, or account-specific behavior.
+
+
+### Hosted request policy context
+
+A trusted host ingress may set `X-Exp-Request-Context` to an opaque value of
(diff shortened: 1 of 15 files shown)


kfallah:
@greptileai review

The public stub now declares supports_request_context and a context-aware host protocol accepted by serve, addressing the review finding. Full Ruff, format and ty checks pass.

kfallah:
Withdrawing this approach per the requested architecture: geographic enforcement will live entirely in Platform/API, using the existing released engine. No package release was published.

PR #867, opened 2026-09-08
providers: preserve Gemini listing page tokens

### LINKED ISSUE #866: Gemini model discovery corrupts opaque pagination tokens
## Summary

Gemini model discovery interpolates `nextPageToken` directly into the next request URL. Because
the token is an opaque provider value, reserved characters such as `+`, `&`, `%`, and `#` are
reinterpreted as URL syntax instead of being preserved.

## Environment

- Surface: authenticated provider model discovery
- Provider: Gemini
- Affected path: `exp/runtime/models/providers/listing.py`
- Affected method: `HttpProviderModelLister._gemini_models`

## Reproduction steps

1. Use the existing scripted JSON transport for Gemini model listing.
2. Return a first page with `nextPageToken` set to `page+2&cursor=%23fragment#tail`.
3. Return an empty second page.
4. Call `list_models` and decode the query string recorded for the second request.

The second request is currently constructed as:

```text
...?pageSize=200&pageToken=page+2&cursor=%23fragment#tail
```

Parsing that URL returns `page 2` for `pageToken`. The remainder is treated as another query
parameter and a fragment, so it is no longer the token Gemini returned.

## Expected

The next request should URL-encode the opaque cursor. Decoding its `pageToken` query parameter
should produce exactly `page+2&cursor=%23fragment#tail`.

## Actual

The cursor is inserted without encoding, which changes or truncates tokens containing reserved
URL characters.

## Impact

An account whose Gemini listing response contains such a cursor cannot reliably enumerate later
model pages. Setup may fail its next request or silently discover only part of the available model
catalog.

sdivyanshu90:
## Why

Gemini model discovery inserted an opaque `nextPageToken` directly into the next request URL.
Reserved characters were interpreted as query or fragment syntax, so a token such as
`page+2&cursor=%23fragment#tail` reached the transport as only `page 2`.

Closes #866.

## What

- Build Gemini model-listing query strings with `urllib.parse.urlencode`.
- Add a transport-level regression test covering `+`, `&`, `%`, `=`, and `#` in one opaque token.
- Decode the recorded request in the test and require an exact token round trip.

## Validation

- `.venv/bin/pytest -q exp/runtime/models/providers/listing_test.py exp/cli/providers/setup_test.py`
  - 41 passed
- `.venv/bin/ruff check .`
  - passed
- `.venv/bin/ruff format --check .`
  - 821 files passed
- `.venv/bin/ty check`
  - passed
- `timeout 180 .venv/bin/pytest -q -x`
  - 108 passed before the unrelated gateway socket test hit the workspace sandbox restriction:
    `PermissionError: [Errno 1] Operation not permitted`

### DIFF
--- a/exp/runtime/models/providers/listing.py
+++ b/exp/runtime/models/providers/listing.py
@@ -11,6 +11,7 @@ import math
 from collections.abc import Mapping, Sequence
 from dataclasses import dataclass
 from typing import Protocol, cast
+from urllib.parse import urlencode
 
 from exp.common.core.artifacts import JsonObject
 from exp.common.models import ReasoningEffort
@@ -204,9 +205,10 @@ class HttpProviderModelLister:
         models = []
         page_token: str | None = None
         for _ in range(_MAXIMUM_GEMINI_PAGES):
-            url = f"{base_url}/models?pageSize=200"
+            query_parameters = {"pageSize": "200"}
             if page_token is not None:
-                url = f"{url}&pageToken={page_token}"
+                query_parameters["pageToken"] = page_token
+            url = f"{base_url}/models?{urlencode(query_parameters)}"
             body = self._read(endpoint, url, headers)
             for entry in _entries(endpoint.provider, body.get("models")):
                 identity = _text(entry.get("name"))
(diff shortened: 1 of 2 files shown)


sdivyanshu90:
Hi @SilenNaihin, could you please review this PR whenever you have some time?

sdivyanshu90:
@kfallah Could you please review this PR whenever you have some time? Thanks

PR #877, opened 2026-09-09
gateway: reserve input tokens from a tokenizer estimate instead of the byte bound

SilenNaihin:
## Problem

`worst_case_input_tokens` returned `len(canonical_json_bytes(request))`: the byte length of the whole canonical request stood in for the input-token reservation. That number fed both the paid worst-case money ceiling (`budgets.maximum_attempt_cost_micro_usd`) and the `reserved_input_tokens` handed to the host's `start_attempt` hook, which the platform compares against free-tier hourly/daily token windows and org token rate limits.

Production (2026-09-09), median `reserved_input_tokens / settled input_tokens` per model:

| model | median ratio |
|---|---:|
| gpt-5.6-luna | 4.69x |
| deepseek-v4-flash | 4.23x |
| gpt-6-astra | 5.38x |
| claude-fable-5.1 | 4.52x |

p90 was much worse on tool-heavy requests (a 17,672-token astra prompt reserved 100,750 input tokens). Because the free-tier gate checks `settled_in_window + reserved_input_tokens <= cap`, a 300k tokens/hour cap bound at roughly 55k real tokens for the prospective request while the dashboard showed a fraction of the allowance used (customer report: org guy-michaely-b42508f6).

## Change

One estimator, both callers. `worst_case_input_tokens` now:

- tokenizes message content, tool-call names/ids/arguments, tool names/descriptions/schemas, the structured-output schema, and the replay envelope (`provider_replay_authority`, so encrypted reasoning, native items, retained raw arguments, and verbatim configs stay counted) in a single `o200k_base` pass (`tiktoken`, cached encoder, warmed at native-bridge construction so a fresh process never pays the table load on its first request);
- adds explicit framing constants (4 tokens per message, 12 per tool, 400 tool-use preamble when any tool is declared) and `INPUT_TOKEN_HEADROOM_PERCENT = 15`;
- reserves planning constants for inline media instead of their base64 length (image 2,048; inline document/video/audio by decoded bytes with floors; URL/handle media by fixed constants), so a 600 KB PNG reserves about 2.4k tokens instead of 800k;
- counts opaque base64 carriers (encrypted reasoning, signatures, sealed content; 256+ chars of base64 alphabet) by decoded length at 4 bytes/token rather than tokenizing the base64 (which would over-count 3x to 4x);
- never double-counts a tool call whose verbatim `raw_arguments` is retained.

`maximum_attempt_cost_micro_usd` takes an optional precomputed `input_tokens` so the ladder walk in `native_accounting` tokenizes once and prices every candidate from that number. Embeddings and image-generation ceilings take `input_tokens` too, so all three surfaces price the same estimate the host reserves.

**Long-context tier decision.** The byte bound guaranteed "estimate >= actual", so an estimate below the tier threshold could never be repriced. The estimate cannot promise that, and the tier reprices the WHOLE request (Gemini doubles rates above 200k). `budgets.LONG_CONTEXT_TIER_MARGIN_PERCENT = 20` treats the tier as reachable from 20% below its threshold: a request estimated at 160k+ against a 200k tier reserves at premium rates and settles at whatever the provider applied. Cost: a one-attempt over-reservation of roughly the tier multiple inside that band. Without it a hard monthly budget could be overdrawn by the same multiple on a threshold-straddling request. Documented in code and in `docs/reference/gateway-architecture.md`.

## Why a tokenizer and not a calibrated byte divisor

Measured o200k bytes/token spans 1.5 (base64) to 7.5 (Cyrillic, Thai); prose 5.7, code 4.9, compact JSON 4.7, pretty JSON 3.8, markdown tables 2.1. No single divisor lands within +5% to +40% across realistic content without undercounting some classes, so a byte heuristic could not meet the calibration bar. `tiktoken` was already in `uv.lock` (via `tinker-cookbook`) and is now a direct runtime dependency (`tiktoken>=0.9,<1`); the crate side is untouched.

## Before/after on the committed fixtures (`attempt_tokens_test.py`)

"Provider count" is an independent rendering (OpenAI cookbook chat framing, compact-JSON tool schemas) tokenized directly, without the estimator's chunking, framing constants, or headroom.

| fixture | provider count | before (bytes) | ratio | after (estimate) | ratio |
|---|---:|---:|---:|---:|---:|
| prose_chat | 199 | 2,070 | 10.40x | 230 | 1.16x |
| coding_agent_15_tools | 47,566 | 243,296 | 5.11x | 54,992 | 1.16x |
| json_heavy | 24,934 | 82,567 | 3.31x | 28,642 | 1.15x |
| multilingual | 2,857 | 14,195 | 4.97x | 3,284 | 1.15x |

Estimator cost on the 47k-token, 15-tool fixture: about 9.7 ms/call (the byte bound was 0.9 ms/call). tiktoken alone encodes 300 KB / 58k tokens in about 10 ms, so the tokenizer, not the walk, is the cost.

## Tests

- New `attempt_tokens_test.py`: calibration over four realistic fixtures (estimate within +5% to +40% of the provider count and never below), framing arithmetic on a hand-counted fixture, tool preamble/schema accounting, media planning constants, opaque carrier decoded-length accounting, raw-argument double-count guard, embeddings/images text, cached encoder plus a loose cost bound.
- `budgets_test.py`: tier margin band semantics, carrier-crosses-threshold rewritten against the estimate, embeddings ceiling parity, precomputed `input_tokens` parity.
- `images_contracts_test.py`: ceiling takes `input_tokens`.
- `uv run pytest exp/runtime/gateway`: 1,048 passed, 3 skipped (serial). `ruff check`, `ruff format --check`, `ty check`: clean.
- Full suite (`-n 4 --dist worksteal -k "not live"`): 3,992 passed, 1 skipped, 2 failed + 1 error, all environmental and unrelated: two release-evidence tests require a checkout with no staged/tracked changes (run pre-commit), one `Config has no attribute cache` from running with `-p no:cacheprovider`. CI runs from a clean checkout with the cache plugin.

## Packaged vocabulary (second commit)

The first push warmed the encoder through tiktoken's registry, which fetches `o200k_base.tiktoken` from OpenAI's public blob on a cold host; the installed-wheel release evidence exited at startup on that path. The wheel now ships the published rank table gzip-compressed (`exp/runtime/gateway/o200k_base.tiktoken.gz`, 1.7 MB; decompressed digest pinned to the same SHA-256 tiktoken itself pins for o200k_base) and `reservation_tokenizer.py` builds the `Encoding` directly from those ranks: no registry, no `tiktoken_ext` plugin discovery, no on-disk cache, no download, no `requests` call. Building the encoder costs about 380 ms once at bind time. A missing or corrupt table raises `ReservationTokenizerError` naming the file at startup.

Verification:
- `reservation_tokenizer_test.py` builds the encoder with every tiktoken download/cache/registry seam monkeypatched to raise and proxies pointed at a dead port, pins the published token ids for a sample, checks the shipped digest and rank count, and covers the corrupt/missing failure messages.
- Installed-wheel release evidence (`exp/tests/release_test.py` driver) run locally against the built wheel in a fresh venv with `HTTPS_PROXY`/`HTTP_PROXY` on a dead port and an empty `TIKTOKEN_CACHE_DIR`: exit 0. The same driver against the first commit's wheel with egress blocked failed exactly at `reservation_encoder()` with a `requests.exceptions.ProxyError` for `openaipublic.blob.core.windows.net`; with egress it passed, so the CI failure was the network path.
- `tiktoken` is declared in `[project].dependencies` and appears in the built wheel METADATA as `Requires-Dist: tiktoken<1,>=0.9`; `REQUIRED_CORE_REQUIREMENTS` in the release gate now includes it.
- The release driver now includes the gateway's stderr and exit code when it exits before listening, so a startup failure is never a bare banner again.

Security scan: the only runtime egress introduced by the first commit (the vocabulary download) is gone; the package fetches nothing at runtime.

## Operational notes

- Settlement is unchanged: actual tokens still replace every reservation.


SilenNaihin:
Security scan note (accepted): the estimate replacing the byte bound is a deliberate product decision (Silen, 2026-09-09): the free-tier reservation must be the same realistic number the paid path uses. Guards: +15% headroom over the o200k count, framing constants per message/tool, media and opaque carriers reserved by planning constants, a 20% long-context tier margin, and settlement always charges the provider's actual usage. The exposure is a bounded one-attempt overdraw on a hard budget for a request whose tokenizer drifts more than the headroom; the previous 4-6x over-reservation was throttling paying customers at a fraction of their allowance. No runtime egress remains (o200k table ships in the wheel, digest-pinned).

PR #948, opened 2026-09-15
fix: accept Azure content-filter annotation frames

ohnoah:
Azure OpenAI can return a successful text completion followed by a content-filter annotation choice with no `delta`. The compatible normalizer currently rejects that documented frame with `OpenAI-compatible delta must be an object`, so a valid response becomes a 502 before trailing usage is consumed.

Treat a missing delta as empty only when both `content_filter_results` and `content_filter_offsets` are objects. Continue through the existing finish-reason handling so `content_filter` remains a refusal. Explicit null/non-object deltas and missing deltas without the annotation shape remain malformed.

Bump the native crate and Python package to `0.3.55`, including lockfiles and the flagship dependency floor, as required by the crate-change CI gate.

Regression coverage exercises annotations before and after the stop chunk, fragmented SSE decoding, trailing usage, content-filter refusal, and malformed frames. Two local HTTP/official SDK tests reproduce the 502 before the patch and return `OK` with 13 input / 1 output tokens after it, through both Chat Completions and Responses.

Reference: [Microsoft Azure content streaming examples](https://learn.microsoft.com/en-us/azure/foundry/openai/concepts/content-streaming).

Validation on commit `bd3a603`:
- [GitHub-hosted gate run](https://github.com/ohnoah/experiential/actions/runs/34942089427): both Linux gate and macOS W16 evidence jobs passed using the repository's unchanged workflow.
- 304 Rust tests, 21 native parity tests, 7 Tinker adapter tests, and 4,198 non-live tests passed (11 skipped).
- Both macOS release-evidence tests passed.
- Full-crate formatting and Clippy with `-D warnings`, plus repository Ruff and ty checks passed.
- The earlier local terminal failures were caused by `TERM=dumb`; all 76 picker tests pass with `TERM=xterm-256color`. Live provider tests and the installed-wheel release test are outside this CI gate; the earlier unrestricted local run hit HTTP 401 and a local Python wheel-tag mismatch on those checks.

Ready for maintainer review. Upstream Actions still requires maintainer approval for this fork contribution. The fork workflow run does not replace required upstream checks; its workflow-dispatch trigger skips the PR-only version comparison, which is satisfied by the synchronized 0.3.54 to 0.3.55 bump.

The fix has been validated with synthetic provider traffic. The hosted service has not been redeployed.


ohnoah:
Additional scope note from the Codex implementation and review:

Microsoft's [streaming documentation](https://learn.microsoft.com/en-us/azure/foundry/openai/concepts/content-streaming) explicitly shows annotation choices without `delta`, including after `finish_reason: "stop"` and with `finish_reason: "content_filter"`. The [generated Azure v1 schema](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/ai/data-plane/OpenAI.v1/azure-v1-v1-generated.json) still requires `delta` in `OpenAI.CreateChatCompletionStreamResponseChoices`. The generic schema alone therefore does not capture the documented annotation shape; this fix is grounded in the streaming examples and an observed Azure response.

Codex deliberately did not implement the following in this PR:

- **A broader parser refactor.** If more supported variants emerge, a small decoder that separately extracts content deltas, annotations, finish reasons, and usage could make the structure clearer. Those fields can coexist, so an exclusive "content or annotation" classification would be too restrictive. The current patch keeps the existing finish/refusal handling shared.
- **Annotation metadata forwarding.** Detailed filter categories/results and offsets are still discarded during normalization. The fix prevents the 502 and preserves terminal refusal and usage behavior, but does not give callers the annotation ranges needed for selective redaction of already-streamed text. A bounded metadata/event contract would be a separate API design change.
- **Blanket acceptance of missing or null deltas.** Explicit `delta: null` and missing-delta choices without both filter metadata objects remain rejected. We have not established additional valid wire examples that justify relaxing those checks.

A fixture corpus across supported Azure API versions and filtering modes would be a useful follow-up. This PR addresses the reproduced failure; it is not a claim of comprehensive Azure schema compatibility.

PR #962, opened 2026-09-15
fix: settle an OpenAI-compatible stream that ends after finish_reason without [DONE] (Azure DeepSeek content filter)

SilenNaihin:
## Problem

Azure AI Foundry's DeepSeek deployments end a **content-filtered** stream with the `finish_reason: "content_filter"` chunk and then close the connection without `data: [DONE]`. The OpenAI-compatible decoder treated `[DONE]` as the only terminal, so the relay's EOF became `provider stream ended without a terminal event` (`malformed_response`, 502 `all_routes_failed` mid-stream) — and because content had already streamed, nothing could fail over.

Production (platform ledger, deepseek-v4-flash on azure_openai, 24h to 2026-09-15 08:00Z): 911 attempts, 843 from one org whose prompts trip Azure's output filter (adult/violent topics), first token at 0.8 s, dead at ~2 s, every request ended `failed`. The worker log's pre-boundary label was `"provider stream ended without a terminal event"` for 37 of 40 malformed events in the last hour.

Staging reproduction (staff-only gateway, same Azure account): a filter-tripping prompt streamed 70 content chunks, then the gateway relayed `{"delta":{"refusal":""}}` — the engine's `RefusalDelta("")` that is emitted **only** on a `finish_reason: content_filter` chunk — followed by the mid-stream error frame; the staging worker logged `{"event":"malformed_response_boundary","reason":"provider stream ended without a terminal event"}` for that request. So the provider did declare its ending; only the sentinel was missing.

## Fix

- `compatible.rs`: the `[DONE]` terminal synthesis moves into `finish_openai_compatible_stream`; `openai_compatible_stream_end` calls it on a clean EOF when a `finish_reason` was already seen (refusal for `content_filter`/`safety`, Incomplete for `length` or a mid-fragment tool cut, Completed otherwise; trailing usage folded). No finish reason → nothing synthesized → `stream_ended` still fails closed.
- `dialects.rs`: `Normalizer::on_stream_end` now returns `Result<Vec<Event>, Failure>` (finishing open tool calls can fail on a syntactically broken argument object, which must stay malformed) and dispatches per dialect: Gemini's existing after-content completion, plus the OpenAI-compatible finish-reason ending.
- `relay.rs` / `drain_stream_fixture`: consume the fallible result (an error routes through `recover_or_fail` / `recover_or_report`, exactly like a decoder failure).
- Native crate `0.3.56 -> 0.3.57` (Cargo.toml, crate pyproject, root lower bound, both locks): the gate requires every crate change to move the version (#964 took 0.3.56 first). **No release cut here**; the platform lead batches releases.

## Tests

- Rust (`compatible_tests.rs`): content_filter finish + EOF → text deltas, refusal delta, `Failed(refusal, content_policy)`; `stop` finish + EOF produces the identical event list to the `[DONE]` twin, usage folded; `length` finish + EOF over an open tool fragment → Incomplete; EOF with **no** finish reason → still `provider stream ended without a terminal event`; a broken argument object under a non-length finish stays malformed. `cargo test --no-default-features`: 309 passed. `cargo fmt --check`, `cargo clippy --all-targets`: clean.
- Python served gateway (`native_openai_compatible_finish_without_done_test.py`, parametrized `content_filter`/`stop`): a loopback provider streams text + finish chunk + usage and closes without `[DONE]`; both endings answer exactly as their `[DONE]` twins (200, content delivered, `finish_reason: "stop"`, usage 13/1) on the non-streaming and the streaming Chat surface, with no mid-stream error frame. `ruff`, `ty`: clean.


SilenNaihin:
Raw Azure evidence (direct probe against the Foundry endpoint, key redacted, 2026-09-15 ~09:40Z, East US 2, `x-ms-rai-invoked: true` on both):

**Streaming, HTTP 200** — the model's stream ends with `finish_reason:"stop"` and a usage chunk, then Azure's RAI layer appends one more frame (different `id`) with `finish_reason:"content_filter"` and closes the connection with **no `data: [DONE]`** — the exact shape this PR settles:
```
data: {"id":"ef69…","object":"chat.completion.chunk","created":1789466242,"model":"DeepSeek-V4-Flash","choices":[{"index":0,"delta":{"reasoning_content":null},"logprobs":null,"finish_reason":"stop","matched_stop":1}]}
data: {"id":"ef69…","object":"chat.completion.chunk","created":1789466242,"model":"DeepSeek-V4-Flash","choices":[],"usage":{"prompt_tokens":88,"total_tokens":195,"completion_tokens":107,"prompt_tokens_details":null,"reasoning_tokens":0}}
data: {"created":1789466242,"id":"chatcmpl-d1b4…","object":"chat.completion.chunk","choices":[{"delta":{},"finish_reason":"content_filter","index":0,"content_filter_results":{"error":{"code":"content_filter","message":"Response content blocked by label 'MultiSeverity_SexualScore'."}}}]}
<EOF>
```

**Non-streaming, HTTP 400 after 35.7 s of generation** — a `chat.completion` body with no top-level `error` (so `rejected_code` reads nothing and it files as generic `invalid_request`); the classification of this shape is a follow-up:
```json
{"id":"chatcmpl-802d…","model":"","choices":[{"index":0,"message":{"role":"assistant","content":""},"finish_reason":"content_filter","content_filter_results":{"error":{"code":"content_filter","message":"Response content blocked by label 'MultiSeverity_ViolenceScore'."}}}],"usage":{"prompt_tokens":55,"total_tokens":55},"created":1789466192,"object":"chat.completion","prompt_filter_results":null}
```

PR #969, opened 2026-09-15
gateway: fold mid-conversation system turns on the Gemini and Bedrock wires instead of refusing the route

SilenNaihin:
## Why

Gemini's `systemInstruction` and Bedrock Converse's `system` carry instructions only outside the turn list, so a `system` turn after conversation start has no positional carrier on those wires, and `streaming_requests.py` refused the WHOLE route for it:

> A system message after conversation start is not supported by this model route. Move the instruction to the leading system prompt or choose a different model.

Coding agents produce exactly that shape on every tool loop (Claude Code's Environment prompt after the first user turn, its `<total_tokens>` reminder after every tool result), and a route needs only ONE Gemini or Bedrock rung anywhere in its waterfall to be refused, even when the Anthropic or OpenAI-compatible rung that would have served it comes first.

## Production evidence (7 days to 2026-09-15, read-only ledger)

| Source | Requests | Orgs |
|---|---|---|
| Pre-admission (`gateway_request_rejections`, code `unsupported_parameter`, param `messages`; the front records no model for 853 of them) | 877 | 22 |
| Admitted, then refused at shaping (`gateway_requests.terminal_error_message`) | 882 | 29 |

By wire: 100% `/v1/chat/completions`. By user agent (pre-admission rows): claude-cli 754, Go-http-client 86, node 17, Python 11, node-fetch 4, Dart 2.

By alias (admitted rows): gemini-3.1-pro-preview-customtools 421, gemini-3.7-flash 136, gemini-3.1-pro-preview 70, gemini-3-flash-preview 54, claude-sonnet-4.5 34, claude-opus-4.6 32, claude-sonnet-4.6 23, minimax-m2.5 22, claude-opus-4.7 21, gemini-3.8-flash 15, deepseek-r1 7, then a tail of gemini-3.x and mistral aliases. Every Gemini alias routes `gemini,openrouter,vertex` (Gemini dialect); every Claude alias routes `anthropic,(azure_openai),bedrock,novita`, so the Anthropic rung would have served the request and the Bedrock rung's presence alone refused it; minimax-m2.5 and grok-4.6 carry a Bedrock rung too.

## What

- `instruction_turns.fold_instruction_turns_after_the_leading_run`: keeps the leading instruction run intact (the builders hoist it one part per message, bytes unchanged) and folds every later plain-text system/developer turn into user text at its position: appended to a preceding text-only user turn, else its own user turn. This is the treatment `anthropic_messages_stream_payload` already gives such a turn.
- `gemini_generate_request` and `converse_body` apply it before hoisting (Gemini accepts consecutive user contents; Converse's `push` merges adjacent user blocks, so alternation holds).
- Admission discloses `messages.system->folded(system_instruction_wire)` (same vocabulary as `messages.system->folded(system_messages_leading_only)` from #956) when a route has a Gemini or Bedrock rung and a turn would move.
- The refusal and its now-unused predicate `mid_conversation_system_present` are removed. No `.rs` change; no crate bump.
- Docs: `docs/reference/gateway-architecture.md`.

## Tests

`instruction_turns_test` (leading run kept, later turns folded, no-op identity), `gemini_requests_test` and `bedrock_requests_test` (two leading parts hoisted, mid-turn instruction merged into the preceding user turn, post-tool reminder merged into the tool-result user message on Converse / its own user content on Gemini), `streaming_requests_test::test_mid_conversation_system_stays_positional_on_capable_wires` rewritten to assert the served payloads and the disclosure on Gemini/Bedrock routes and none on an Anthropic-only route, plus a native-bridge end-to-end on a configured Gemini rung (Claude Code's Chat shape → `systemInstruction` carries only the leading text, `contents` carry the folded turn in place, `ignored_parameters` holds the disclosure).

`ruff check`, `ruff format --check`, `ty check` clean; full `pytest` run locally.

## Production proof (after the roll)

The class must go to zero on both sources while the same aliases keep completing:

```bash
.local/pq.sh "select date_trunc('hour', created_at) h, count(*) from gateway_request_rejections where created_at > now() - interval '48 hours' and message ilike '%system message after conversation start%' group by 1 order by 1"
.local/pq.sh "select date_trunc('hour', accepted_at) h, count(*) from gateway_requests where accepted_at > now() - interval '48 hours' and terminal_error_message ilike '%system message after conversation start%' group by 1 order by 1"
```

Probe (production probe key, on the lead's go): the Chat body from platform#1882's proof comment against `gemini-3.7-flash` and against `claude-sonnet-4.6`, expecting 200 and `x-experiential-ignored-parameters` containing `messages.system->folded(system_instruction_wire)` when the serving rung's route carries a Gemini/Bedrock rung.


SilenNaihin:
Production proof 2026-09-15 ~19:55Z on experiential 0.7.78 (platform e155c524a, 8 workers on …cc13ace12): claude-cli-shaped Chat probe with a system turn at index 3 → gemini-3.7-flash served on vertex (request-17f3d219cfa6415b84db2ac694e5fadf, completed, text "done") and claude-sonnet-4.6 served on bedrock (request-862a33e6d6034381a7cc32177ae47d8f, completed, text "done"), both 200 with `x-experiential-ignored-parameters` = [`messages.system->folded(system_instruction_wire)`]. Pre-admission class: 915 rows/22 orgs in the 7 d before the roll, 0 after; admitted-then-refused class: 920/27 orgs before, 0 after. No provider-side ordering rejection on any gemini/vertex/bedrock rung since the roll.

PR #1056, opened 2026-09-20
providers: paginate Anthropic model discovery

### LINKED ISSUE #1055: Anthropic model discovery silently stops after the first page
## Summary

Authenticated Anthropic model discovery reads only the first `/v1/models` response and ignores the documented `has_more` and `last_id` pagination fields. Accounts with more than one page silently omit callable models from setup and provider synchronization.

Classification: P2 priority, S difficulty, intermediate experience, API surface, low risk, high confidence, good first issue.

## Environment

- Current canonical `main`
- Python 3.12.13
- Linux

## Reproduction steps

Configure the existing scripted listing transport with two successful Anthropic responses:

```json
{"data":[{"id":"claude-opus-5"}],"has_more":true,"last_id":"claude-opus-5"}
{"data":[{"id":"claude-haiku-4-5"}],"has_more":false,"last_id":"claude-haiku-4-5"}
```

Then call `HttpProviderModelLister.list_models` for an Anthropic endpoint.

The result contains only `claude-opus-5`, and the second scripted response remains unread.

## Expected

Discovery should follow `last_id` through the `after_id` query parameter while `has_more` is true, then return models from every page. Pagination must be bounded, encode the cursor as an opaque query value, reject missing or repeated cursors, and fail loudly rather than silently truncate at its safety cap.

Anthropic's current Models API reference documents a default page size of 20, a maximum requested size of 1000, and `has_more` plus `last_id` for forward pagination:

https://platform.claude.com/docs/en/api/models/list

## Actual

`HttpProviderModelLister._anthropic_models` performs one request to `/models`, reads only `data`, and returns immediately. It does not inspect the pagination envelope.

## Impact

Provider setup and synchronization can present an incomplete authenticated model catalog without warning. Users may be unable to select a model their Anthropic account can call, even though authentication and discovery otherwise succeed. This is silent functional truncation, not credential loss or a provider outage.

## Suggested resolution

- Request the provider's maximum documented page size to minimize calls.
- Follow `last_id` as an encoded `after_id` while `has_more` is true.
- Enforce a finite page ceiling and fail if the provider still reports more data at that ceiling.
- Reject `has_more` with the wrong type, a missing cursor, or a repeated cursor.
- Add deterministic coverage for a two-page result, malformed pagination, cursor repetition, and cap exhaustion.

This does not change authentication, model metadata interpretation, provider selection, or catalog merging.

## Duplicate search

Immediately before filing, open and closed canonical issues and pull requests were searched for `anthropic pagination`, `has_more last_id models`, `after_id anthropic models`, `_anthropic_models`, `Anthropic model discovery pagination`, and `model listing anthropic`. No matching report or fix was found. Pull request #783 paginates the gateway batch API, not authenticated provider-model discovery. Pull request #457 introduced the discovery surface without adding Anthropic pagination.

sdivyanshu90:
## Why

Anthropic's Models API is cursor-paginated. Model discovery currently reads only the first response, so a catalog with `has_more: true` silently omits every later model.

## What changed

- request the largest documented Anthropic model page
- follow `last_id` through the `after_id` query parameter
- encode opaque cursors before adding them to the request URL
- reject malformed or repeated pagination cursors instead of returning a partial catalog
- stop after a finite ten-page ceiling and fail loudly if the provider still reports more data
- cover multi-page discovery, cursor encoding, malformed metadata, cursor cycles, and cap exhaustion

## Validation

- `uv run --no-sync pytest -q exp/runtime/models/providers/listing_test.py` (21 passed)
- `uv run --no-sync pytest -q exp/cli/providers exp/cli/gateway/setup_test.py exp/runtime/models/providers/listing_test.py` (180 passed)
- `uv run ruff check .`
- `uv run ruff format --check .` (926 files already formatted)
- `PYTHONPATH=exp/runtime/gateway/native uv run --no-sync ty check`

The explicit `PYTHONPATH` makes the checked-in native stub available because the locally installed extension predates the current upstream source.

## Non-goals

This does not change provider selection, capability inference, or any Rust code.

Closes #1055


sdivyanshu90:
@SilenNaihin @kfallah Could you please review this PR whenever you have some time? Thanks

